PR #14, opened 2021-02-20
wagtail apiv2

Adrianbarros:
wagtail installed and apiv2 set up


isedwards:
@Adrianbarros the changes are not yet ready to review because the commit contains your `venv` directory and also other non-code files like `*.pyc` files (5,000+ files).

First, on your computer, you will need to undo your last commit using `git reset --soft HEAD~1`

Then, if you type `git status` you will see thousands of files we don't want to include in the changes that we push.

You then need to add a suitable `.gitignore` file to the top level directory in the repository so that the files we don't want to share are ignored. For now, we might as well use the one that is being used for IoGT v1 (see [here](https://github.com/unicef/molo-iogt/blob/develop/.gitignore)).

The `.gitignore` file linked above has at least two useful entries...
```
*.py[cod]
venv
```
The first line shown ignores `*.pyc`, `*.pyo` and `*.pyd` files (which are not source code files). The second line shown will ignore your entire `venv` directory.

We'll probably need to add an extra line with `.vscode` to the file to exclude VSCode settings.

When you push your changes you won't be able to push to your existing `wagtailinstall` branch unless you either delete it first (by editing your personal branches [here](https://github.com/Adrianbarros/iogt/branches)). Or you have to use the `-f` command line switch to force your changes to replace what you had pushed before (I'd recommend avoiding this option for now. In general you want to avoid using the force option to remove old history).

PR #40, opened 2021-03-05
More detail in readme

istride:


README.md
```diff
@@ -13,19 +13,34 @@ Inline with the latest Wagtail [Long Term Support release](https://github.com/wa
...
 ```
-mkdir iogt-dev
-cd iogt-dev
 git clone https://github.com/unicef/iogt.git
+cd iogt
+```
+
+Create an [isolated Python environment][1], into which we will install the project's dependencies. This is highly recommended so that the requirements of this project don't interfere with any other Python environments you may have in your system, including the global Python environment.
+```
 python3 -m venv venv
```
isedwards, line 23: I usually create the `venv` directory outside of the git repository just to make it simpler when doing searches, e.g. `grep -R text_to_find *`.

However, it's okay to create it inside the repository since the `.gitignore` file will ignore the contents.  Which do you think is better for contributors (your way certainly looks tidier for the instructions!)?

README.md
```diff
@@ -4,7 +4,7 @@ Internet of Good Things ([IoGT](https://www.unicef.org/innovation/IoGT)) is deve
 
 The development uses the Python programming language, building on the [Django](https://www.djangoproject.com/) server-side web framework combined with the [Wagtail](https://wagtail.io/) content management system.
 
-Inline with the latest Wagtail [Long Term Support release](https://github.com/wagtail/wagtail/wiki/Release-schedule), IoGT 2.x supports:
+In line with the latest Wagtail [Long Term Support release](https://github.com/wagtail/wagtail/wiki/Release-schedule), IoGT 2.x supports:
```
isedwards, line 7: Could change my markdown links so that they use the same style as yours (which will keep line lenghts shorter)?

isedwards (changes requested):
Minor suggestions

README.md
istride: ah, I misunderstood your initial intention. I agree that the venv should be outside the project dir, though perhaps we ought not to be too prescriptive about where it should go.

README.md
```diff
@@ -13,19 +13,39 @@ Inline with the latest Wagtail [Long Term Support release](https://github.com/wa
...
-mkdir iogt-dev
-cd iogt-dev
 git clone https://github.com/unicef/iogt.git
-python3 -m venv venv
-source ./venv/bin/activate
-
 cd iogt
+```
+
+Create an [isolated Python environment][1], into which we will install the project's dependencies. This is highly recommended so that the requirements of this project don't interfere with any other Python environments you may have in your system, including the global Python environment. We aso recommend that you create a new environment outside of the project directory.
```
isedwards, line 21: Typo in final sentence ("also")

README.md
```diff
@@ -13,19 +13,39 @@ Inline with the latest Wagtail [Long Term Support release](https://github.com/wa
...
-python3 -m venv venv
-source ./venv/bin/activate
-
 cd iogt
+```
+
+Create an [isolated Python environment][1], into which we will install the project's dependencies. This is highly recommended so that the requirements of this project don't interfere with any other Python environments you may have in your system, including the global Python environment. We aso recommend that you create a new environment outside of the project directory. 
+```
+python3 -m venv <path_to_venv_dir>
+source ./venv/bin/activate
```
isedwards, line 24: Add custom path to second venv command:
`source <path_to_venv_dir>/bin/activate`

isedwards (changes requested):
Just needs a change to the second venv path

isedwards (approved):

PR #43, opened 2021-03-15
basic home page / section / article structure

istride:
Opening pull request to get early feedback.


isedwards (approved):
@istride, this is exactly what I wanted to see in an initial implementation.

The only comment I have is that I don't think we should merge the migrations file (e.g. `migrations/0003_article_section.py`) into the main branch yet... We should wait until we finalise the `Article` model.

This will allow us to locally delete any early migrations that are proposed whilst the model is still a bit volatile and instead commit a simpler set of migrations to the repository when everything has settled down.

Issue #19, opened 2021-02-22
As CMS editor, I can add and remove footer pages


lauriemarkle:
Footer pages may include terms & conditions, about IoGT, etc and are pages that appear on the site footer. They should be fully editable pages, as articles.

dannyparsons:
Footers implemented here #96 as fully editable pages

PR #45, opened 2021-03-17
add editable footer snippet

istride:
Related to issue #19 . A fully editable footer snippet that allows changes to navigation links, logos and the "essential" links to privacy policy, terms of use, etc. All links are to pages within the site.

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -30,21 +30,21 @@ class Article(Page):
 
 @register_snippet
 class Footer(models.Model):
-    title = models.CharField(max_length=255, blank=True)
+    title = models.CharField(max_length=255)
     logos = StreamField([
-        ('image', ImageChooserBlock())
-    ], null=True)
+        ('image', ImageChooserBlock(required=False))
+    ], blank=True)
     navigation = StreamField([
         ('link_group', blocks.StructBlock([
             ('title', blocks.CharBlock()),
             ('links', blocks.StreamBlock([
                 ('page', blocks.PageChooserBlock())
             ]))
-        ]))
-    ], null=True)
+        ], required=False))
+    ], blank=True)
     essential = StreamField([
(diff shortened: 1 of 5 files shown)


home/models.py
```diff
@@ -26,3 +27,31 @@ class Article(Page):
     content_panels = Page.content_panels + [
         StreamFieldPanel('body')
     ]
+
+@register_snippet
+class Footer(models.Model):
+    title = models.CharField(max_length=255, blank=True)
+    logos = StreamField([
+        ('image', ImageChooserBlock())
+    ], null=True)
```
isedwards, line 36: The `logos`, `navigation` and `essential` fields are currently all mandatory required fields (despite your `null=True`)...

The confusion is discussed [here](https://github.com/wagtail/wagtail/issues/4306).

In addition to `null=True`, `blank=True` and `required=True`... both `StreamField` and `StreamBlock` have `min_num` and `max_num` arguments (discussed [here](https://github.com/wagtail/wagtail/issues/5175)).

isedwards (changes requested):
Great start of getting footers added.  I've added a couple of initial comments.

home/templatetags/home_tags.py
```diff
@@ -0,0 +1,11 @@
+from django import template
+from home.models import Footer
+
+register = template.Library()
+
+@register.inclusion_tag('home/tags/footer.html', takes_context=True)
+def footer(context):
+    return {
+        'footer': Footer.objects.all()[0],
```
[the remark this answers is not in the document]
istride: Yes, worth tidying this up. I expect we'll come up with a better mechanism for choosing the correct footer later on.

home/models.py
istride: quite confusing. i've attempted to make only `title` and `essential` required fields, the rest are optional.

PR #47, opened 2021-03-23
Article pages

istride:
Creating a template to render article pages, copying CSS from PWA project, small changes to Article model to support the design of the page.


isedwards (approved):
Usually I'd suggest removing the hard-coded `.survey-secton` form and the `.comments-section` div, but I can see the argument that we should leave them in for the example demo.

PR #54, opened 2021-04-14
Test of automated migration of content from molo iogt database

michaelclapham:
Just experimenting with reading from the old database, and automatically populating section pages in the new Wagtail setup.﻿


isedwards:
@istride could you take a look and see how this fits with the work that you've started?

Michael took the approach that I originally suggested of reading directly from the legacy database and automating the process of `Article` creation.

istride:
Yes, this is will pretty much what I plan to do, but in the form of a management command.

PR #61, opened 2021-05-04
app engine deploy

istride:
All the changes I made to deploy to App Engine. Ideally, I'd like to keep all deployment-related information out of this repository, but can't think of a quick solution. Something to work on later, perhaps.


isedwards (approved):
Looks good, thank you for getting all this done.

My only comment is - I'd probably move the dependencies that are specific to deploying on Google App Engine into a separate `requirements-gae.txt` (that includes the main requirements file at the top):

```
-r requirements.txt

google-cloud-storage
google-cloud-secret-manager`
```

PR #62, opened 2021-05-05
Added ElasticSearch for searching Articles

### LINKED ISSUE #57: Basic search
Users should be able to search for content.

Users can enter search terms in input box in the site header.
Users are shown a page of results.
Search index should use Elasticsearch as a backend.

ehmadzubair:
Fixes #57 
- Added ElasticSearch backend
- Updated index for Articles
- Wired the search page to the search header
- Updated Readme
- Updated Search UI to match Home UI
- Created Template Tag for Article List
- Added `docker-compose.yml`


ehmadzubair:
@istride Not happy with the way the results look right now. Is there a design that we need to replicate for the results? Also, is that part of Issue #57 ?

istride:
@ehmadzubair The search results page could be in the same style as the home page.

ehmadzubair:
@istride Updated the PR. Will dockerize this in a separate PR. Can you review this whenever you get time?

ehmadzubair:
@istride Added `docker-compose.yml`

istride:
@ehmadzubair I've had a look, and I think it's a good start, but needs some polish.

It took me a while to get things up and running because your instructions (for Docker Compose, in particular) in the readme assume that a database is already set up and migrated, and that an elasticsearch index has been created. I suggest making this clear in the readme. I also suggest creating a settings file just for use with Docker Compose, as opposed to using local.py.

Would you tell me what your method for testing the pagination is?

The site is supposed to have a responsive design, so please check that the search results page looks ok at all breakpoints.

There is a search input or icon at the top of every page, as part of the design, so another search box on the results page seems unnecessary. The title "Search", is also unnecessary and should be replaced with something like, "Search results for <search term>".

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: '3'
+
+services:
+  django:
+    container_name: iogt-dg01
```
istride, line 5: I would just use `name` instead of `container_name`. Docker Compose will add the project name and numbering all by itself.

ehmadzubair: I think `name` is supported by the `network` and `volume` blocks only. Am i missing something here?

ehmadzubair:
@istride Good points about the docker compose setup. I was assuming the basic setup instructions shouldn't be repeated but you're right. I've updated the readme and added the settings file too.

> Would you tell me what your method for testing the pagination is?

For pagination, I just created more articles and tested the results (decreasing the page size also works).

> The site is supposed to have a responsive design, so please check that the search results page looks ok at all breakpoints.
I noticed the Home page isn't formatted for all screen sizes either. Since I've moved it to a template tag, this would be reflected across the website. Can you tell me the specific requirement for how articles should look on all sizes? For example, the number of columns on each size.

ehmadzubair:
Also, I'm a little rusty on the css end - it's been a while. Do we have css help on the project or is that an expectation from me? If it's the latter, I can just brush up on it and be back in the game.

docker-compose.yml
istride: No, I was mistaken. What I was actually suggesting was to leave `container_name` unspecified i.e. remove it. Docker Compose should automatically name containers with the following scheme <project name>_<service name>_<container number> e.g. iogt_django_1.

istride:
We will eventually have some help with the styling, but we all have to maintain it, so it would be good to brush up on CSS. It does look better without the extra search bar, at least. Since there isn't very strong direction on the styling front, I think we'll leave it as it is for now.

The docker compose setup is still a bit clunky, but it works and it can be improved over time.

When I created new articles, I had to manually run the update process to add them to the search index. This obviously can't be the solution in production. Ideally, there would be an automatic process that adds articles to the index when they are published, and removes them if an article is ever deleted or unpublished. What do you have in mind for that?

ehmadzubair:
There is an auto_index flag that can be used to achieve just that. However, I feel that too frequent (atomic) updates in the index can be computationally expensive if the dataset is large.

I think we can make a more informed decision when we have a sense of how much data needs to be indexed. Adding a cron and/or manual indexing is also an option but it really depends on the specifics of the index.

dannyparsons:
I think it's necessary that articles are searchable as soon as they are added, and removed from the index once deleted/unpublished. I don't think the extra computation should be a concern.

ehmadzubair:
Just looked up the documentation and there's an `ATOMIC_REBUILD` flag that rebuilds the index without downtime. Setting `AUTO_UPDATE` and `ATOMIC_REBUILD` to `True`

ehmadzubair:
The only concern I have now is that the index will be rebuilt on every change. This could impact the performance whenever we're editing the content. It could be a problem when the dataset becomes very large.

Thoughts? @istride @dannyparsons

istride:
> The only concern I have now is that the index will be rebuilt on every change. This could impact the performance whenever we're editing the content. It could be a problem when the dataset becomes very large.

I was hoping for a more surgical update of the index that only affects a single article, but this does not seem to be available at the moment. I think auto update is ok in development, but we might have to think of an alternative for production.I don't expect any further work on this as part of this task though.

I have no other objections, so I think this pull request is ready to be merged. I'll do this fairly soon. Thanks for your help, @ehmadzubair .

PR #66, opened 2021-05-11
Iogt comment system

MySecondLanguage:
Following task is implemented for comment

- Commenting can be enabled or disabled on any page.
- Users can post comments.
- Users can reply to comments.
- Users can 'like' comments.
- Moderators can review, filter and export comments.
- Moderators can provide custom or canned responses to comments.
- Community moderators can flag inappropriate comments.
- Moderators should be alerted to flagged comments.


istride:
I have encountered a few issues:

* Not able to reply to a comment; clicking on the 'reply' link does nothing
* I can 'like' and 'flag' my own comments; don't think that should be allowed
* When a comment is created it would be less confusing to reload the page with the newly created content in view (same with liking, flagging and replying)

I am also having trouble working out how I would do the content management/moderation features you mentioned:

> Moderators can review, filter and export comments.
> Moderators can provide custom or canned responses to comments.
> Moderators should be alerted to flagged comments.

Would you be able to write some brief instructions on how you expect an admin/moderator to achieve those things?

MySecondLanguage:
> * When a comment is created it would be less confusing to reload the page with the newly created content in view (same with liking, flagging and replying)

As you said the system will be no-js, i had not option except reloading. If you allow me to use ajax, then i will implement like, comment, flag with ajax, so we can avoid reload the page.

Issue #27, opened 2021-02-24
As a front end visitor, I can create and edit my account


lauriemarkle:

dannyparsons:
How should "Forgot my password/PIN" work? In the current sites, you enter your username and it says there is a security question to answer but nothing appears. When you click "Forgot PIN" there is an error.

![image](https://user-images.githubusercontent.com/11214914/118482084-2c8f2000-b70c-11eb-9921-5100bd7e8710.png)

Some options:
- A security question could be added to the registration and used to reset the PIN. 
- Security question/email could be optional fields and PIN reset is only possible if one of these has been given. 
- PIN reset is not available

@lauriemarkle Can we discuss this in our sessions next week?

lauriemarkle:
Sounds like a good idea. Let's walk through the whole user registration
process. I think it's 80% of the way there.

ehmadzubair:
@ChrisMarsh82 Can you set up a meeting to go over with (Ideally with Laurie)?

ChrisMarsh82:
linked to #446

PR #69, opened 2021-05-15
Feat/user registration

### LINKED ISSUE #55: User account registration / management
Visitors of IoGT sites should be able to sign-up for an account to allow for personalisation and community involvement.

- Visitors should be able to create a user account for themselves.
- As part of the sign-up process visitors may be asked a set of questions, which may be different for each site, and may be optional or mandatory.
- Users should be able to change details of their account (name, username, password)
- Users should be able to initiate a password reset process.

ehmadzubair:
Fixes #55 #27 

Things to discuss before merging this pull request:
- Finalise the Email templates for Password Reset and Email confirmation (if applicable)
- Do we need to ask the user for First Name and Last Name on Signup?
- Are there any settings that need to be configured for django-allauth? See documentation [here](https://django-allauth.readthedocs.io/en/latest/configuration.html)


ehmadzubair:
@istride @dannyparsons The first chunk of the user registration is complete in this PR.

Here is a list of the functionality that has been added:
- User Sign In/Sign Out
- My Profile section (accessible from the link in the header)
- Edit My Profile screen (accessible from My Profile screen)
- Change Password feature (accessible from My Profile section)
- Forgot Password feature (accessible from the Login screen)
- Rudimentary CSS
- Added support for django's messaging framework

dannyparsons:
@ehmadzubair Thanks. Will test again once the fields are updated. Note display name will be a single field, not first name and surname.

Also, the forgot password feature will need some more thought as email won't be supplied by default. Will get back to you on that.

ehmadzubair:
@istride This is probably something that we can discuss in our call today. Customising the User model is a major change in any Django project and is a one way operation. Most of these decisions can only be done at the start of the project.

There are 3 ways to add `full_name` field to the User:
1. Use the existing field `first_name` on the `User` model as full name. This would help preserve the default `django.contrib.auth.models.User` class. This is semantically misleading but we can write docs around this to make sure everyone understands this. This would also involve overriding the form field to read `full_name`
2. Add a `full_name` field to the `users.Profile` model. This is slightly counter intuitive but has the same advantages as the option above.
3. Create a new User class and update the fields. This is a one-way operation and can't be undone. If we foresee frequent changes to the user model in the future, we should go with this option. This will also require some extra work on `django-allauth`'s implementation.

ehmadzubair:
More research: Moving to a custom user model mid-project is risky and we should replace the `auth.User` with our custom model before the project goes into production.

[Here](https://code.djangoproject.com/ticket/25313) is a ticket that talks about change to custom user mid-project

ehmadzubair:
Items left in this ticket:
- Merge feat/custom-user-model
- Add display_name field
- Update Signup form to receive display_name

dannyparsons:
@ehmadzubair The site isn't in production yet and we haven't migrated any data yet so I don't see an issue with this but @istride can confirm.

We don't need a `full_name` field we only need: `username`, `display_name` and `PIN/password`.

isedwards:
I agree with following the recommended best practice and enabling a custom user model from the outset:

https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#using-a-custom-user-model-when-starting-a-project

I'm not sure what the implications of this change will be for the demo site (where we want to keep the content that has been entered). It's worth testing on a local copy to check the easiest way to perform the update and keep existing content.

ehmadzubair:
@dannyparsons @istride 

Just completed the user registration feature. Before testing the PR, you will have to delete your database and set up the project data again.

Here is the list of features that were done yesterday and today:
- User registration with custom user model
- Added Display Name and a 4 digit password (although we'll have to find a hack to allow numbers only)
- Restrict access to private pages if the user hasn't submitted the post registration survey
- Added Accept terms & conditions checkbox
- Did a bunch of changes in All-auth's configuration (without breaking any support for its functionality)

Hope this is at a decent spot now where we can merge it and plug-in the post registration form some time later (maybe with the surveys feature) 🤞🏻

istride:
After conversation with @ehmadzubair , we identified a number of issues to be addressed:

- review naming of `IogtUser` and `IogtAccountAdapter` with view to removing `Iogt` prefix
- removing all django_allauth templates that we do not have a use for at the moment (we'll add them when necessary)
- in Wagtail Admin's user view - fixing the name where it shows as "None None"; promoting username as the main identifier and showing Display Name if it's available
- understand why there are multiple 'group' and 'permissions' tables and remove any that are not necessary

ehmadzubair:
@istride Just got done with making all the changes

> review naming of IogtUser and IogtAccountAdapter with view to removing Iogt prefix

Done

> removing all django_allauth templates that we do not have a use for at the moment (we'll add them when necessary)

Done

> in Wagtail Admin's user view - fixing the name where it shows as "None None"; promoting username as the main identifier and showing Display Name if it's available

Done - however there's a very minor UI bug (actually a feature) with the highlighted `Name` column. that's because the default ordering on the view is done on `Name`. Overriding the view was proving to be a heavy lift so I didn't spend much time on it

> understand why there are multiple 'group' and 'permissions' tables and remove any that are not necessary

It's actually very funny how we both missed it. Here's what they do:
- auth_group is for the names of the group
- auth_group_permissions is the m2m between permissions and groups
- auth_permissions is for listing all the permissions on the platform
- togt_users_user_groups is the m2m between iogt_users_user to auth_group
- iogt_users_user_user_permissions is the m2m between iogt_users_user and auth_permission


Screenshot 2021-05-21 at 11 23 21 PM

ehmadzubair:
@istride did the following changes:
- Removed Iogt prefixes
- Deleted usused templates
- Update the column placement in Users IndexView
- Overrode the Create/Edit User forms in Wagtail Admin
- Changed Sign In to Sign Up on header
- Got rid of the user pagination warning
- Renamed `has_accepted_terms_and_conditions` to `terms_accepted`
- Added comment for Surveys in the custom decorator
- Cleaned migrations (you'll have to delete the sqlite database again)

Issue #60, opened 2021-05-01
Draft Sections and Articles are shown in navigation


isedwards:
Sections and Articles that have not been published are currently being shown on the home page and in the navigation (but a 404 error is given when the article is clicked since the Article has not been published).

sheralim012:
@ChrisMarsh82 this is done and you can close this.

geoo89:
Confirmed this is fixed.

PR #71, opened 2021-05-19
stop to rendering unpublished articles

djangodeveloper96:
@istride  Fixed to rendering the unpublished articles.

### DIFF
--- a/home/templatetags/home_tags.py
+++ b/home/templatetags/home_tags.py
@@ -1,16 +1,17 @@
 from django import template
-from home.models import Footer
+import home.models as model
 
 register = template.Library()
 
 @register.inclusion_tag('home/tags/footer.html', takes_context=True)
 def footer(context):
     return {
-        'footer': Footer.objects.first(),
+        'footer': model.Footer.objects.first(),
         'request': context['request'],
     }
 
 
 @register.inclusion_tag('home/tags/articles_list.html')
 def render_articles_list(articles):
-    return {'articles': articles}
+    live_articles = articles.filter(live=True)
+    return {'articles': live_articles}
--- a/home/templatetags/home_tags.py
+++ b/home/templatetags/home_tags.py
@@ -1,5 +1,6 @@
 from django import template
 from home.models import Footer
+from home.views import render_live_articles
 
 register = template.Library()
 
@@ -13,5 +14,4 @@ def footer(context):
 
 @register.inclusion_tag('home/tags/articles_list.html')
 def render_articles_list(articles):
-    live_articles = articles.filter(live=True)
-    return {'articles': live_articles}
+    return {'articles': render_live_articles(articles)}
--- a/home/templatetags/home_tags.py
+++ b/home/templatetags/home_tags.py
@@ -1,12 +1,12 @@
 from django import template
-import home.models as model
+from home.models import Footer
 
 register = template.Library()
 
 @register.inclusion_tag('home/tags/footer.html', takes_context=True)
 def footer(context):
     return {
-        'footer': model.Footer.objects.first(),
+        'footer': Footer.objects.first(),
         'request': context['request'],
     }
 
--- a/home/views.py
+++ b/home/views.py
@@ -1,0 +1,5 @@
+
+
+def render_live_articles(articles):
+    live_articles = articles.filter(live=True)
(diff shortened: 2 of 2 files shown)


home/templatetags/home_tags.py
```diff
@@ -1,16 +1,17 @@
...
-from home.models import Footer
+import home.models as model
 
 register = template.Library()
 
 @register.inclusion_tag('home/tags/footer.html', takes_context=True)
 def footer(context):
     return {
-        'footer': Footer.objects.first(),
+        'footer': model.Footer.objects.first(),
```
ehmadzubair, line 9: I think it's better to leave this as-is. `Footer.objects.first()`

home/templatetags/home_tags.py
```diff
@@ -1,16 +1,17 @@
...
         'request': context['request'],
     }
 
 
 @register.inclusion_tag('home/tags/articles_list.html')
 def render_articles_list(articles):
-    return {'articles': articles}
+    live_articles = articles.filter(live=True)
```
ehmadzubair, line 16: I think we should alter filter these articles from the view vs the template tag. I haven't tested this, but [https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html](the documentation) seems to suggest that we can filter it by `Page.objects.live()...`

home/templatetags/home_tags.py
isedwards: @ehmadzubair could you add the reason why you would leave this as it was? There doesn't seem to be any significant difference between the two.

home/templatetags/home_tags.py
djangodeveloper96: It works, I tested it before pushing.

`template tag` has 4 loops.
[https://github.com/unicef/iogt/blob/336d94d7fc7ee4426962219aeb90ee9829c78f28/home/templates/home/tags/articles_list.html#L4]
[https://github.com/unicef/iogt/blob/336d94d7fc7ee4426962219aeb90ee9829c78f28/home/templates/home/tags/articles_list.html#L22]
[https://github.com/unicef/iogt/blob/336d94d7fc7ee4426962219aeb90ee9829c78f28/home/templates/home/tags/articles_list.html#L41]
[https://github.com/unicef/iogt/blob/336d94d7fc7ee4426962219aeb90ee9829c78f28/home/templates/home/tags/articles_list.html#L74]

In this case, I thought it's better to filter from view instead of template tag.

home/templatetags/home_tags.py
djangodeveloper96: done

ehmadzubair: In my experience, a couple of reasons:
- I think importing models as a whole module can create confusions in future when importing from multiple modules. For example, `import users.models as users_models`. This might result in confusing imports in multiple files. In another file, there could be `import users.models as models`.
- Importing the exact model class (`from home.models import model1, model2, model3`) is the de-facto import style in most of the tutorials, documentation and open source code.

Does this make sense?

home/templatetags/home_tags.py
ehmadzubair: I meant it's more 'standard' to filter the articles in the view. Business logic should not spill over into the templates. The templates should be as `dumb` as possible

djangodeveloper96: So can we keep current updates?

ehmadzubair: I would say we should still filter the articles in the views and not in template tags

istride: I agree with @ehmadzubair. This is a rather simple case, so it does not seem to make much difference where the query goes, however, I suspect there may be more things to consider in the future than just whether articles are live or not, such as which order the articles should be in, and perhaps handling other items than articles.

We should also consider how easy it is to test, given that the home page will be more complicated later on. It would also be good to include a test that ensures that only live articles are present on the home page, @djangodeveloper96 .

djangodeveloper96:
This is Roland
https://github.com/unicef/iogt/issues/5#issuecomment-845835657
Could anyone reply for this comment?

dannyparsons:
fixes #60

djangodeveloper96:
> fixes #60

pushed my updates.

PR #77, opened 2021-05-27
Added support for video and audio

### LINKED ISSUE #8: As a CMS editor, I can upload multimedia files on the site for users to stream from the front end.
Multi-media files can be uploaded as MP3 & MP4 format. These files can then be accessed when editing articles and loaded as feature content. When file uploads exceed 6MB a warning sign appears that the format may not be accessible on low end devices. 

(bonus points: can the MP4 function as a banner on the homepage?)

sheralim012:
Fixes #8 

Following changes are done in this PR:
- Added audio and video support in Admin
- Added media block in Article
- Added Media content on the Article detail screen


sheralim012:
@istride Can you go through this whenever you have time?

cc: @dannyparsons

istride:
Hi @sheralim012 . Thanks for the PR, which I think is ok.

In issue #8 , there is a requirement to show a warning message, in the admin interface, if a media file exceeds a certain size limit.

>When file uploads exceed 6MB a warning sign appears that the format may not be accessible on low end devices.

Do you have a solution for that?

sheralim012:
Hey @istride, 
As we are using `wagtailmedia` 3rd-party package to upload video & audio and it provides a simple way to override its form and include additional validation and raise appropriate **errors** but there is no simple way to show **warning** message and to show a warning message, we need to override views from `wagtailmedia` (add & edit both cases).

Currently, I'm looking for a way to handle this using javascript in templates for uploading media.

dannyparsons:
@sheralim012 Would it be easier to implement a file size limit (determined in site settings) which gave an error if a file exceeds the limit?

sheralim012:
@dannyparsons yes, it's simple to give an error if the file size limit exceeds.

dannyparsons:
Ok, please investigate the warning option. If this becomes too complicated, we might check if the simple limit with error is sufficient.

dannyparsons:
This looks good. The media block in articles works really well too. A couple of minor changes:

- There should be an option to choose which Collection to upload to.
- Can duration be removed? Not a big issue if not but it's not used. It's good that it's optional anyway.

dannyparsons:
Image uploading also needs the Collection option (could be addressed in a separate PR though)

sheralim012:
@dannyparsons file size exceeds warning is addressed.
also, the collection option already exists and will show while file uploading if any collection exists (default is root collection and not shown).
and to remove duration option, we would have to override `wagtailmedia` form.

dannyparsons:
Thanks Sher. Collections work as you mentioned. Ok I think it's not worth overriding just to remove duration, so let's leave it as is.

I uploaded two videos of 10MB and 17MB but I didn't get any warning. Do I need to do something before this shows?

sheralim012:
@dannyparsons there are a couple of things that you could do,
- make sure to pull the latest code
- restart the server
- clear browser cache or try incognito mode

If this doesn't work then let me know which browser you are using. Thanks.

iogt/static/js/iogt.js
```diff
@@ -0,0 +1,13 @@
+function validateFileUpload(fileInput) {
+  if (!fileInput.files || !fileInput.files[0])
+    return true;
+  else {
+    var file = fileInput.files[0];
+    if (file.size >= 9437184)
+      return confirm('The file you have uploaded exceeds 9mb. This will prohibit access to the file in a low ' +
+          'bandwidth setting, may restrict feature phone access, or violate your mobile network operator ' +
+          'agreements. To reduce file size, try resizing and compressing your file.');
```
dannyparsons, line 9:
```suggestion
          'agreements. To reduce file size, try resizing and compressing your file. Do you want to contnue?');
```

dannyparsons:
Got it thanks, had to clear browser cache. This works well. I just edited the warning message slightly.

@istride This works as needed in the CMS Editor, over to you for final review.

iogt/static/js/iogt.js
```diff
@@ -0,0 +1,13 @@
+function validateFileUpload(fileInput) {
+  if (!fileInput.files || !fileInput.files[0])
+    return true;
+  else {
+    var file = fileInput.files[0];
+    if (file.size >= 9437184)
```
ehmadzubair, line 6: I think this size and the user facing size mentioned in the string below need to be configurable from the backend. @istride can confirm.

dannyparsons: Yes, that could be useful. @sheralim012 you could start on the Site Settings issue #32 and add this as an option there. But with default 9MB still. It may be best to do this in a separate PR, so that this can be merged as-is.

mwrobel-merixstudio: even if we want to use some Site Settings in near future I will strongly suggest to already implement this as const in iogt/iogt/settings/base.py and mark this place with # TODO comment

istride: At the moment there is a magic number in the code that needs some explanation. Perhaps assigning that magic number to a constant would help.

Making the warning limit configurable from the same place that other things are configured would make it obvious where  the limit can be changed. I had quite a bit of difficulty getting the warning message to show up in the first place - at first it only worked in incognito/private browsing, but eventually worked outside of that mode, though I'm not sure exactly what I did to make it work. If the warning limit was changed someday, would it require a similar effort from CMS admins to clear their browser cache, in order for the updated limit to take effect?

@sheralim012

sheralim012: @istride as Danny mentioned, the file size threshold will be addressed in #32 and I have already created a PR for that and added the threshold there and once this gets merged I'll use that setting here.

If the server is already running and we add/change JS code in templates, sometimes the server doesn't reflect the latest changes and we have to restart the server, and sometimes the browser doesn't reflect the latest changes and we have to clear the browser cache.

dannyparsons: I don't expect this limit to be changed frequently, if at all, as most sites would use the default. So having to clear the browser cache when changing shouldn't be much of an issue. We should document this though for CMS Editors.

PR #79, opened 2021-05-28
Fix - Group Permission Restriction

### LINKED ISSUE #78: Adding group restriction to Page(s) results in an infinite loop
This issue is the result of an open bug https://github.com/wagtail/wagtail/issues/4297 in the Wagtail project. It's very strange how this hasn't been closed yet.

Thankfully, there is a workaround on the same link for this.

ehmadzubair:
Fixes #78 

We'll still need to add a template for this 403 page to look 'decent'. Not spending time on it right now.


istride:
@ehmadzubair my understanding is that you will be adding some test cases to this PR

ehmadzubair:
Yes, that is correct. Right after I complete comments

ehmadzubair:
@istride I have updated this PR with the Test Cases

iogt/templates/header.html
```diff
@@ -36,7 +36,7 @@
         </div>
         <div class="header-middle">
             <a class="logo-small" href='/'>
-                <img src="{% static "/images/iogt-logo-two-lines.png" %}"/>
+                <img src="{% static "images/iogt-logo-two-lines.png" %}"/>
```
ehmadzubair, line 39: This was causing the test suite to not run.

ehmadzubair:
@isedwards If you can review/merge this please.

istride (changes requested):
@ehmadzubair would you add some instructions to the README about how to run the tests. Perhaps add a target to the Makefile.

I'm having trouble running the tests. Firstly, some issue with static files, for which, I ran `./manage.py collectstatic` to resolve. Secondly, I have an issue with the Django Debug Toolbar, which I have not found a fix for, as follows:

```
django.urls.exceptions.NoReverseMatch: 'djdt' is not a registered namespace
```

ehmadzubair:
@istride I've added separate settings for running tests and both the issues have been resolved (not snoozed). I've also added the readme steps to run the tests

istride (approved):

Issue #37, opened 2021-02-24
Community Moderator user group allows enrolled front-end users to assist with comment moderation.


lauriemarkle:
They will have the ability to delete or respond to users with canned responses (developed by CMS editors on backend). In their profile, they can view the total number of comments moderated (total and in the last week).

ehmadzubair:
This functionality is done already. The only thing probably left in this is to create a user group for community moderators. @ChrisMarsh82 can you close this ticket and create one for Community Moderator user group which only has access to the comments section on the admin side bar?

ChrisMarsh82:
Community moderator - no access to the back end - moderate comments only
Moderator - as is
Editor - as is

ehmadzubair:
@ChrisMarsh82 I think the requirements for this are a little conflicting. Laurie mentioned that moderators can respond with canned responses (developed by editors on backend). The also have the ability to view their moderated comments. Would it make sense to give them limited access to the admin panel?

cbunicef:
The challenge is that the admin panel is not feature phone-friendly unfortunately.


[Icon  Description automatically generated]Christopher Brooks
Internet of Good Things<https://unicef.sharepoint.com/teams/PD-IGT/> – Global Lead
Email: ***@***.******@***.***>

ChrisMarsh82:
Community Moderator needs to do admin through the front end, this should be feature phone friendly.

geoo89:
A first draft of this has been implemented, ready for external review.

There are two levels of permissions here:
- Deleting comments in the front-end: enabled via: Object Permissions > Comment > Delete
- Add canned responses in the front-end: other Permissions > Can Moderate Comments

A community moderator group can be created and have permissions for either of both of these, without having access to the admin panel. 
Additional permissions are needed for viewing/moderating comments via the admin panel, e.g. for a full Moderator group).

Any user can report comments (no permissions needed).

cbunicef:
Peer moderators should not have the power to "delete" comments, but rather simultaneously "hide" and "flag".

PR #81, opened 2021-06-01
Feat/comments

### LINKED ISSUE #33: As a front-end user I can post comments when I'm logged in and reply to other users' comments. I can like comments.
This feature already exists, but the UX is not great: comments should appear align left v align right so that users can see what's a comment v what's a reply. 

New feature: liking comments. Comments on smart phones and above can be viewed by most likes or most recent. 

In v3.0 and beyond, users should be able to upvote / downvote and like comments.

### LINKED ISSUE #34: As a front-end user, I can report comments that violate community policies
Community members can self-moderate by reporting inappropriate comments. 
Reported comments appear as an alert in the CMS.

### LINKED ISSUE #35: As a comment moderator, I can review comments, export comments, filter comments and provide custom or canned responses

### LINKED ISSUE #56: Comments
Registered users should be allowed to comment on content.

- Commenting can be enabled or disabled on any page.
- Users can post comments.
- Users can reply to comments.
- Users can 'like' comments.
- Moderators can review, filter and export comments.
- Moderators can provide custom or canned responses to comments.
- Community moderators can flag inappropriate comments.
- Moderators should be alerted to flagged comments.

ehmadzubair:
This PR fixes #56 

Here's a list of the features done as part of this:

> - Commenting can be enabled or disabled on any page. 

This is done for the Page and its comments as well. Disallowing comments hides both the "New Comment" form and the "Reply" links on existing comments

> - Users can post comments

Go to Article screen and scroll to the bottom part of the page. Make sure that "Allow Comments" is set to true on Admin > Pages > Article > Settings

> - Users can 'like' comments

Done

> - Moderators can review, filter and export comments. 

See Admin > Comments > All Comments. Look for button on hovering the comments.

> - Moderators can provide custom or canned responses to comments.

See Admin > All Comments > Reply button on a comment

> - Community moderators can flag inappropriate comments.

Everybody can flag comments for now but only moderators can delete comments. Since flagging is also a requirement for frontend users, the same feature is shared between them. Thoughts @dannyparsons @istride?

> - Moderators should be alerted to flagged comments.

Discussed with @dannyparsons that this can be replaced with a filter. The filter can be found here "Admin > Comments > All Comments > Flagged Filter"

> - Fix #33 

All the requirements for #33 are complete except for comment sorting. This PR was becoming very lengthy, so I decided to merge this first and then do the minor changes in another PR

> - Fix #34 

Flagging is the same as reporting. 

> - Fix #35 

Done. Add canned responses from Admin > Comments > Canned Responses

> - #37 

This will have to be done in a separate PR

## Testing Instructions
There are no special steps required for testing. Just running the migrations and running the app should show up the comments box and comments list on the frontend. The admin panel should have a new sections named "Comments" on the sidebar

Note: Created #86 for issues that are not completed in this


comments/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
istride, line 3: would rather omit this file, if there are no tests

home/templates/home/article.html
```diff
@@ -1,8 +1,11 @@
 {% extends "base.html" %}
-{% load static wagtailcore_tags wagtailimages_tags %}
+{% load static wagtailcore_tags wagtailimages_tags comments comments_xtd %}
 
 {% block extra_css %}
-<link rel="stylesheet" type="text/css" href="{% static 'css/article.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static 'css/article.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static 'css/comments.css' %}">
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.3.1/css/all.css" integrity="sha384-mzrmE5qonljUremFsqc01SB46JvROS7bZs3IO2EmfFsd15uHvIt+Y8vEf7N7fWAU" crossorigin="anonymous">
```
istride, line 7: I think we'll have to insist that all assets are served by IoGT to ensure that users don't incur data charges. The addition of font awesome may be temporary, but I would prefer that we always make sure that we host everything ourselves at all times.

home/models.py
```diff
@@ -90,11 +91,18 @@ def get_paragraph_values(self):
         paragraph_values = self._get_child_block_values('paragraph')
         return '\n'.join(paragraph_values)
 
+    def get_absolute_url(self):
```
istride, line 94: not sure why this is necessary

comments/models.py
```diff
@@ -0,0 +1,19 @@
+from django.db import models
+from django_comments_xtd.models import XtdComment
+
+
+class AllowCommentsModelMixin(models.Model):
```
istride, line 5: would consider renaming this to `CommentsMixin` or `CommentableMixin`.

what happens if other attributes are required in this class e.g. time-restricted commenting?

comments/views.py
```diff
@@ -0,0 +1,69 @@
...
+from django_comments.views.comments import post_comment
+from django_comments_xtd import get_form
+from django_comments_xtd.models import XtdComment
+from django.utils.translation import ugettext as _
+
+from comments.forms import CommentForm, AdminCommentForm
+from comments.models import CannedResponse
+
+
+def update(request, comment_pk, action):
```
istride, line 17: My code smell senses are tingling, but I am not able to offer a viable alternative. Thinking from a RESTful perspective, I would want each 'action' here to be handled by a separate endpoint and this would eliminate the 'if-elif' structure here.

If no change is made, should there be some tests to catch possible regressions in the future if someone has to edit this code?

home/models.py
```diff
@@ -90,11 +91,18 @@ def get_paragraph_values(self):
...
+        return self.get_url()
+
     content_panels = Page.content_panels + [
         ImageChooserPanel('lead_image'),
         StreamFieldPanel('body')
     ]
 
+    settings_panels = Page.settings_panels + [
+        FieldPanel('allow_comments')
```
istride, line 103: My preference would be for the panel to be titled 'Comments', and for there to be a checkbox labelled 'Enabled' within the panel.

istride (changes requested):
A few more things that I noticed:

- every interaction with the comments system, as a user, redirects me to 'example.com'
- not a fan of asking for confirmation when liking comments
- i can still see the old comments template on article pages (is this a case of clearing browser cache?)
- if comments are switched off, I think there should be no mention of comments anywhere
- as an admin editing a comment, there a lot of attributes that can be changed, which perhaps, should not be allowed to be changed, such as 'content_type', 'object_id' - admins could get into real trouble by fiddling with these.

comments/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
```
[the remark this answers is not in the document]
ehmadzubair: This is auto-generated by Django. Will do

comments/models.py
ehmadzubair: Shouldn't 'other required attributes' be set in the setting panels?

comments/views.py
ehmadzubair: This code has been 'copied' from `wagtail_comments_xtd`. Will discuss in the call

home/models.py
ehmadzubair: The `comments_xtd` library uses the `et_absolute_url` function of the model to redirect to after every successful action.

iogt/settings/profanity_settings.py
```diff
@@ -0,0 +1,2 @@
+COMMENTS_ALLOW_PROFANITIES = False
+PROFANITIES_LIST = ('crap', 'hell')
```
ehmadzubair, line 2: Discuss with @istride

home/models.py
ehmadzubair: Move this to the Mixin

ehmadzubair:
@istride 

> every interaction with the comments system, as a user, redirects me to 'example.com'

I couldn't reproduce this. 

> not a fan of asking for confirmation when liking comments

Will figure out how to get rid of this (default behaviour of comments-xtd)

> i can still see the old comments template on article pages (is this a case of clearing browser cache?)

This was pushed later in `02ee0de`. You shoulndn't see it now.

> if comments are switched off, I think there should be no mention of comments anywhere

@dannyparsons and I had discussed that setting allow comments to false would only stop new responses. Do we want to hide all the comments with the same flag OR create a new flag for this?

> as an admin editing a comment, there a lot of attributes that can be changed, which perhaps, should not be allowed to be changed, such as 'content_type', 'object_id' - admins could get into real trouble by fiddling with these.

Good catch. Will limit the fields

dannyparsons:
On disabling comments, I think we should keep existing comments visible if they have been made. That's what done in v1 I think and I haven't seen a request to change this. I notice that on articles that have never had comments there's no mention of comments https://sa.goodinternet.org/sections/climate-change/solutions/renewable-energy/ so good to keep that functionality if possible.

ehmadzubair:
> On disabling comments, I think we should keep existing comments visible if they have been made. That's what done in v1 I think and I haven't seen a request to change this. Comments can be deleted if needed separately.
> 
> I notice that on articles that have never had comments there's no mention of comments https://sa.goodinternet.org/sections/climate-change/solutions/renewable-energy/ so good to keep that functionality if possible.

Got it. I've changed the functionality to be exactly the same

ehmadzubair:
@istride All the changes that were requested have been done. I've also incorporated all the previous feedback except for the test cases. I've added that to #86

istride:
@ehmadzubair would you limit font awesome to the bare minimum necessary to make your feature work, or just leave it out altogether. I'm not sure what the new design will be like and whether it will require font awesome or not. I certainly don't want to commit nearly 2000 files, only to take them away a few days later, just for the sake of a few icons.

ehmadzubair:
> @ehmadzubair would you limit font awesome to the bare minimum necessary to make your feature work, or just leave it out altogether. I'm not sure what the new design will be like and whether it will require font awesome or not. I certainly don't want to commit nearly 2000 files, only to take them away a few days later, just for the sake of a few icons.

Yes, will do. I'll do the latter.

ehmadzubair:
@isedwards, @istride and I went through this PR in detail and there were two changes that were pending yesterday. I have made both of them. Can you merge this soon? (it has already caused a lot of diverging migration trees)

cc: @dannyparsons

Issue #12, opened 2021-02-18
As a CMS editor, I can modify the site's logo with an image upload from my gallery


lauriemarkle:
Image size specifications should be noted in the CMS as an information box / hover .

dannyparsons:
Logo is chosen in Site Settings similarly to v1:

![image](https://user-images.githubusercontent.com/11214914/122749102-d8380c80-d284-11eb-9b89-c2ba0c1dfe2f.png)

sheralim012:
@ChrisMarsh82 this isn't implemented yet.

PR #84, opened 2021-06-01
Feat/site settings

### LINKED ISSUE #32: As CMS Admin I can manage site settings
Existing Features to Keep: 
- Logo editing
- Checkbox to show only translated pages
- GA and FB Analytics Links
- Content Rotation Settings
- Social Media Footer Page Links
- Social Media Article Sharing Buttons

Unsure about Existing Features / Need to Investigate: 
- Article tag settings
- Service directory API settings

Features to Add: 
- Color Scheme

sheralim012:
Fixed #32 

- Register wagtail site settings app in installed apps
- Added new site settings
- Added wagtail site settings context processor
- Added wagtail site settings in base template
- Implement media file size threshold site setting


home/migrations/0005_sitesettings.py
```diff
@@ -0,0 +1,50 @@
...
+        ('wagtailcore', '0059_apply_collection_ordering'),
+        ('home', '0004_auto_20210409_1243'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='SiteSettings',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('show_only_translated_pages', models.BooleanField(default=False, help_text='When selecting this option, untranslated pages will not be visible to the front end user when they viewing a child language of the site')),
```
dannyparsons, line 23:
```suggestion
                ('show_only_translated_pages', models.BooleanField(default=False, help_text='When selecting this option, untranslated pages will not be visible to the front end user when viewing a child language of the site')),
```

dannyparsons:
Thanks for this. I'm wondering how this is going to function for multi-sites, as most of these settings need to be different for each site. We should also discuss this with @istride.

sheralim012:
Screenshot 2021-06-02 at 4 28 27 PM
with more than one site configured from **menu>sites**, site setting automatically show dropdown for multisite.

home/migrations/0005_sitesettings.py
sheralim012: change addressed in new push.

sheralim012:
fixed migration file, so before pulling new code run the following commands
```
make ssh
cd /app/
python manage.py migrate home 0004
```
then you can pull new code and run new migrations

dannyparsons:
Thanks @sheralim012 Good to know Site Settings work with multi-sites.

Is there also a way we can restrict access to Site Settings for users with different permissions?

sheralim012:
@dannyparsons you can see all permission in groups from **settings>groups><group-name>** and seems like there isn't a setting for that.

dannyparsons:
Ok that's something we need to discuss. I can see there's a Permission under Object Permissions to be able to change Site Settings. But we need to also be able to define which Site Settings a user can change.

Possibly having a "Sites Permission" section when defining a group that says which sites the group should have access to and then using this in different places to display correctly e.g. Site Settings, Comments.
We should discuss with @istride about possible implementation.

sheralim012:
* go to **Settings>Sites**, create new sites (double check domain and port and prefix localhost so you don't have to add any entry in **/etc/hosts** file)
Screenshot 2021-06-08 at 4 12 55 PM
* go to **Pages>Pages**, create new homepages for each site
Screenshot 2021-06-08 at 4 13 09 PM
* go to **Settings>Groups**, create new groups for each site
Screenshot 2021-06-08 at 4 13 28 PM
* specify permission for these groups
Screenshot 2021-06-08 at 4 13 54 PM
Screenshot 2021-06-08 at 4 14 05 PM
Screenshot 2021-06-08 at 4 14 26 PM
* go to **Settings>Users**, create new users for each site and assign appropriate roles or groups
Screenshot 2021-06-08 at 4 17 19 PM
Screenshot 2021-06-08 at 4 17 37 PM

**Note:** 
* now go to each site wagtail admin (http://site1.localhost:8000/admin/) and login, user can only see things for which access has been granted
* superuser would be able to see all sites' site settings, each site user would be able to see its own site settings.

sheralim012:
@dannyparsons @istride  https://github.com/unicef/iogt/issues/32#issuecomment-853747165 for this we can create callables that will return GA and FBA id from django settings(settings will get it from environment variable) and for each site they would be automatically filled and down the road we still be able to change according to specific site if need be.

home/wagtail_hooks.py
```diff
@@ -0,0 +1,32 @@
...
+    if request.user.is_superuser:
+        menu_items[:] = [item for item in menu_items if item.name != 'site-settings']
+
+
+@hooks.register('construct_settings_menu')
+def hide_site_settings_in_settings_menu(request, menu_items):
+    if request.user.is_superuser:
+        return
+
+    menu_items[:] = [item for item in menu_items if item.name != 'site-settings']
```
ehmadzubair, line 32: Have to be sure why we're making the shallow copy here and if it's necessary altogether?

sheralim012:
Hey @istride, could you please go through this PR.

home/models.py
```diff
@@ -219,3 +220,172 @@ class Footer(models.Model):
 
     def __str__(self):
         return self.title
+
+
+@register_setting
+class SiteSettings(BaseSetting):
```
istride, line 226: @dannyparsons Do we really need all of these settings? For example, where will the content rotation settings be used?

@sheralim012 I suggest, and if Danny agrees, that we cull this list of settings to ones which we are actively using right now in v2 and leave it to others to add settings here when features are implemented.

iogt/static/js/iogt.js
```diff
@@ -1,12 +1,13 @@
...
+function validateFileUpload(fileInput, file_size_threshold) {
   if (!fileInput.files || !fileInput.files[0])
     return true;
   else {
     var file = fileInput.files[0];
-    if (file.size >= 9437184)
-      return confirm('The file you have uploaded exceeds 9mb. This will prohibit access to the file in a low ' +
-          'bandwidth setting, may restrict feature phone access, or violate your mobile network operator ' +
-          'agreements. To reduce file size, try resizing and compressing your file. Do you want to contnue?');
+    if (Math.round(file.size / 1024 / 1024) >= file_size_threshold)
```
istride, line 6: I would prefer to store `file_size_threshold` as an integer number of bytes and then format/parse to/from a human-readable form.

istride (changes requested):

home/models.py
sheralim012: @dannyparsons let me know which fields to keep and which fields to let go.

iogt/static/js/iogt.js
sheralim012: sure

home/models.py
dannyparsons: I've asked a question on the issue about the need for content rotation. I think all the others have a clear use. I believe all the GA and FB analytics will still be used.

dannyparsons: Content rotation isn't in use in v1 so let's remove this.

sheralim012: content rotation fields are removed.

iogt/static/js/iogt.js
sheralim012: done

sheralim012:
fixed migration file, so before pulling new code run the following commands
```
make ssh
cd /app/
python manage.py migrate home 0004
```
then you can pull new code and run new migrations

sheralim012:
Hey @istride, changes addressed.

dannyparsons:
fixes #12

sheralim012:
Resolve conflicts, removes unused imports and fixed migrations, so before pulling new code run the following commands
```
make ssh
cd /app/
python manage.py migrate home 0004
```
then you can pull new code and run new migrations

dannyparsons:
Agreed with @isedwards to merge

PR #92, opened 2021-06-05
feat/info about leaving a free-data environment (IU-12)

mikowhy:
### Solution

- create a basic transition page with info about leaving a free-data environment for: 
  - `RichTextField`s + `wagtail_hooks.py` 
  - `URLField` banners


isedwards:
@mikowhy - can you clarify how the `.env` file will work?

We're intending to have a single server instance that serves multiple sites from the same instance - it looks like the `.env` approach would require a different server instance to be running for each of the sites?

(Also - not sure why all of the single quotes `'` have changed to double quotes `"`. Is this something that your editor has done automatically?  It's affecting a lot of lines of code and isn't part of the proposed solution. It would be good to revert these lines)

mikowhy:
Hi @isedwards,

> can you clarify how the `.env` file will work?
> We're intending to have a single server instance that serves multiple sites from the same instance - it looks like the `.env` approach would require a different server instance to be running for each of the sites?

I was not fully aware of this solution, thinking that the website for each country is located on a different server instance.  Which variable was used to identify the country so far?

> (Also - not sure why all of the single quotes `'` have changed to double quotes `"`. Is this something that your editor has done automatically? 

this is part of [black formatter](https://pypi.org/project/black/) for django which is widely used by us. We discussed how to lint our code and mentioned about that solution.

> It's affecting a lot of lines of code and isn't part of the proposed solution. 

only lines in files which I edited

> It would be good to revert these lines

ok, of course it can be reverted. Which linters (black, flake8, sort...) do you propose for us to have the same rules as you use?   I think it's a helpful solution to automate formatting

isedwards:
Thank you @mikowhy - it looks like I probably missed the conversation about Black.

@istride do you think updating *existing code* to follow Black formatting is a priority for the first release? If so, can we move this into a seperate issue and we can keep PRs for formatting changes separate from functional chages to code. We should be following PEP8 (and using flake8) but I'd encourage these formatting changes (for existing code) to also be in separate PRs where it's clear no functional changes are being made.

(I'm very happy for new code, that is making functional changes, to follow both Black and PEP8)

@mikowhy at the moment it's difficult to be certain what changes are proposed to the settings files (other than formatting changes), if you could revert the formatting changes in this PR and just keep the functional changes then that would be great.  Also, would you be happy to move the addition of django debug tool bar into a sepate PR?

I was assuming we could identify whether we were leaving the current site by checking the `request` data when the content is submitted to find the host that the user is currently at.  Is `request` available at this point?

isedwards:
@lauriemarkle do you know whether a user on free basics can visit any *.goodinternet.org site for free, or is their free access limited by their network provider to just the subdomain for their country (e.g. ke.goodinternet.org)?

mikowhy:
>Also, would you be happy to move the addition of django debug tool bar into a sepate PR?

@isedwards -> https://github.com/unicef/iogt/pull/93

istride:
> do you think updating _existing code_ to follow Black formatting is a priority for the first release?

@isedwards not a priority, but I did agree to a linter being used as long as it was introduced in an incremental way.

> (I'm very happy for new code, that is making functional changes, to follow both Black and PEP8)

I think this is a sensible suggestion, and if the linter ends up complicating things at this stage I'd rather introduce it later once there is less pressure to complete by a deadline.

lauriemarkle:
They can only visit their country site. Free Basics has an admin portal, I
decide in each country where Free Basics sits what URL I want to direct
them to. So, for example, in Southeast Asia five Free Basics countries
navigate to the regional IoGT site (sa.goodinternet.org), but Pakistan kept
their own site (pk.goodinternet.org).

Friendly reminder about this feature:
Every site needs its own IP address and the notification needs to be
translatable and included in the additional text that Alex sends over for
site translation.

mwrobel-merixstudio (approved):

home/models.py
```diff
@@ -182,6 +184,14 @@ class BannerPage(Page):
         FieldPanel('external_link'),
     ]
 
+    def final_external_link(self):
```
ehmadzubair, line 187: Can we make this a `@property`?

mwrobel-merixstudio (approved):
looks ok

home/models.py
mikowhy: done!

mikowhy:
@dannyparsons 
- add `@property` as @ehmadzubair pointed
- and there is ok from @mwrobel-merixstudio 

Something more should be added / changed?

ehmadzubair:
Looks good to me. It would be great to have a test case around the redirection.

mikowhy:
tests for links in `RichText` fields added

README.md
```diff
@@ -97,7 +97,7 @@ make setup
 
 Create a super user account for administration purposes.
 ```
-docker-compose run django python app/manage.py createsuperuser
+docker-compose run django python manage.py createsuperuser
```
ehmadzubair, line 100: I think we need to update the workdir in the Dockerfile.dev for this to work

home/tests/test_links_in_page/test_external_links.py
```diff
@@ -18,8 +23,25 @@ def create_published_article_to_root(self, body) -> Article:
...
+            ("paragraph", RichText('<p>Lorem ipsum <a href="http://aynthing.com">external link in middle</a> dolor sit amet</p>')),
+            ("paragraph", RichText('<p>Lorem dolor sit amet <a href="http://aynthing.com">external link at the end</a></p>')),
+        ]
+
+        article = self.create_published_article_to_root(body=body)
         response = self.client.get(article.get_url())
 
+        assert response.status_code == status.HTTP_200_OK
+
+        html_parser = BeautifulSoup(response.content, "html.parser")
```
ehmadzubair, line 51: @mwrobel-merixstudio How do you feel about using `assertTempalteUsed` and assert-ing on `response.context`?

Here's a [https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Testing#views](link) on how that can be done.

requirements.dev.txt
```diff
@@ -2,4 +2,4 @@
 pytest
 pytest-django
 Faker
-factory-boy
```
ehmadzubair: Is there a reason why we removed factory-boy from here?

mwrobel-merixstudio: we decide not to use it in this PR 
you can add it later if it will be needed

README.md
mwrobel-merixstudio: it already is ;) https://github.com/unicef/iogt/pull/92/files#diff-86930c95a19b82f7e64a962a0053d44e855824813019b3698eae4917a90cdcac

I added some general improvements to docker configuration

home/tests/test_links_in_page/test_external_links.py
mwrobel-merixstudio: I don't think this is part of this test - we want to test only parsing content, and which template we are using template isn't relevant in this test / code implementation. this test should:
- prepare proper Article
- make sure number of links source of Article is same as number in rendered Article
- make sure only links we want to be parsed are parsed
- make sure we parse every link that lead outside Wagtail

if there is possibility to use different templates to render Article - yes, then we should add template assert. @mikowhy is this change of template from admin panel possible?

ehmadzubair: Got it. Thanks

isedwards:
Add this to v2.1 milestone to tidy up the current parallel solutions:
- Rich text input (#92)
- Middleware (#167)

PR #96, opened 2021-06-07
Feat/implement footer

### LINKED ISSUE #95: Update footers implementation
- Footers should be implemented as Page and should inherit from Article
- FooterIndexPage will hold all footers
- Footers will be orderable

sheralim012:
Fixes #95 

- Added Footer Page on wagtail admin
- Added Footer Index Page on wagtail admin
- Footer Index Page will hold all Footer Pages
- Footer Listing Wagtail Admin View will default to orderable view


dannyparsons:
This looks good to me. The Footer Index Page is a sensible way of managing footers. In particular, having ordering by default is really nice and a big improvement from v1 that has been requested.

@istride @isedwards over to you for review.

home/views.py
```diff
@@ -0,0 +1,16 @@
+from django.http import HttpResponseRedirect
+from wagtail.admin.views.pages.listing import index as wagtailadmin_explore
+
+from home.models import FooterIndexPage
+
+
+def override_wagtailadmin_explore_default_ordering(request, parent_page_id):
```
istride, line 7: Not a big fan of this solution. I think there may be a better approach by using the [construct-explorer-page-queryset wagtail hook][1].

I would also suggest that perhaps it's not such a big deal for manual ordering (ordering=ord) to be enabled by default, but it is important for the ordering to be by 'path' regardless.

[1]: https://docs.wagtail.io/en/stable/reference/hooks.html#construct-explorer-page-queryset

iogt/urls.py
```diff
@@ -1,15 +1,18 @@
...
 from wagtail.core import urls as wagtail_urls
 from wagtail.documents import urls as wagtaildocs_urls
 
+from home.views import override_wagtailadmin_explore_default_ordering
+
+
 urlpatterns = [
     path('django-admin/', admin.site.urls),
-
+    re_path(r'^admin/pages/(\d+)/$', override_wagtailadmin_explore_default_ordering),
```
istride, line 15: If wagtail hooks can be used, then we don't need to change `urlpatterns`.

istride (changes requested):

home/views.py
sheralim012: yes, sure if enabling ordering by default is not a big then it far better to use wagtail hook. Changes addressed.

iogt/urls.py
sheralim012: No need for this if we use wagtail hook. Changes addressed

home/views.py
sheralim012: also, should we do this just for footer pages or for all the pages?

sheralim012:
resolve conflicts and fixed migration file, so before pulling new code run the following commands
```
make ssh
cd /app/
python manage.py migrate home 0004
```
then you can pull new code and run new migrations

home/templates/home/tags/footer.html
```diff
@@ -1,27 +1,11 @@
 {% load static wagtailcore_tags wagtailimages_tags %}
 <footer class="footer">
     <nav>
-        {% for block in footer.navigation %}
         <div class="links">
```
ehmadzubair, line 4: We should probably move this template one level up since this is not going to be a template tag anymore

sheralim012: as we don't pass data to it from context but fetch data before rendering and include it as a self-contained tag so we don't need to move it.

sheralim012:
Hey @istride, could you please go through this PR.

home/views.py
istride: Perhaps it should be for other pages too, but just footer pages is fine for now.

istride (approved):

PR #106, opened 2021-06-10
Enhancements to User data and Groups on Admin

### LINKED ISSUE #31: As site administrator I can view the total number of users registered, users by group and export a list of users with key profile information
Note that currently you can export users but it goes to a predetermined email address which is now defunct. Users should be able to download to their device.

ehmadzubair:
Fixes #31 

- Added User count to Groups Listing
- Added Export Item in Admin
- Added UserAdmin with export functionality

## Testing Instructions
- The total number of users can be seen on the **Export > Users Data** screen (under the search bar)
- Users can be filtered by groups, date joined, staff status and active status (other filters can be added too)
- Users can be exported in the xlsx, csv format from **Export > Users Data** screen


ehmadzubair:
@dannyparsons I had initially added the user count to the groups admin page but that segregated the information in different pages.

All of the changes for this ticket should be available on the **Export > Users Data** screen

@isedwards can you review this when you have time?

ehmadzubair:
@istride Can you review/merge this when you have time?

iogt_users/templates/modeladmin/iogt_users/user/index.html
```diff
@@ -0,0 +1,106 @@
+{% extends "wagtailadmin/base.html" %}
```
istride, line 1: Is this template copied from somewhere? Is it possible to extend instead and only change the header block to add the user count? Otherwise, it's hard to see where the customisation is, and potentially we are opening ourselves up to maintaining the whole template when we don't have to.

iogt_users/wagtail_hooks.py
```diff
@@ -0,0 +1,29 @@
+from django.contrib.auth import get_user_model
+from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register, ModelAdminGroup
+
+from iogt_users.filters import GroupsFilter
+
+
+class UsersExportAdmin(ModelAdmin):
+    model = get_user_model()
+    menu_label = 'Users Data'
```
istride, line 9: Change label to simply say 'Users'.

istride (changes requested):

isedwards:
@ehmadzubair - could you confirm whether [iogt_users/templates/modeladmin/iogt_users/user/index.html](https://github.com/unicef/iogt/pull/106/files/210e63a0e64a702015bd984e7b8f0a9f719d899c#diff-5a4b959b34dd363ab0741012ca248c9364ebedbe1131e3bd68d1584d32fc4b8b) contains duplicated text from another template?

Let's merge and revisit this if necessary.

PR #110, opened 2021-06-14
Added caching via workbox

Kishokanth-netizen:
User can now click on download website button to cache the content that we deem fit. Backend devs will support this by providing access to the content that will be cached


mwrobel-merixstudio (approved):

home/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+class TestView(TemplateView):
+    template_name = "home/test.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data()
+        context["id"] = "lorem ipsum"
+        return context
+        
+class ServiceWorkerView(TemplateView):
```
istride, line 13: I would prefer it if the service worker was just a normal static JS file that fetches the list of URLs to pre-cache from a server endpoint.

iogt/templates/base.html
```diff
@@ -41,9 +44,14 @@
             </div>
         </div>
 
-        {# Global javascript #}
-        <script type="text/javascript" src="{% static 'js/iogt.js' %}"></script>
-        {% block extra_js %}{% endblock %}
+         {# Global javascript #}    
 
+        <script>
+        function cache() {
```
istride, line 50: Is there a good reason why this function is inline rather than in a separate JS file?

I think it would be better to always register the service worker automatically, if possible, and activate caching as a separate step. We would only want to show the option to "Download website" if the service worker is successfully registered.

istride (changes requested):
Not sure how important this is, but I can see a warning, in Firefox, regarding the web manifest that says, "The start URL is outside the scope, so the scope is invalid".

Kishokanth-netizen:
> Not sure how important this is, but I can see a warning, in Firefox, regarding the web manifest that says, "The start URL is outside the scope, so the scope is invalid".

resolved issue

iogt/templates/sw.js
```diff
@@ -0,0 +1,15 @@
+importScripts('https://storage.googleapis.com/workbox-cdn/releases/6.1.5/workbox-sw.js');
```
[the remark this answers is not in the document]
Kishokanth-netizen: Noted. Workbox is now self-hosted.

home/views.py
Kishokanth-netizen: Yep, I agree and this was one of the first approaches I tried. Unfortunately, I ran into some issues. 

The service worker must essentially be defined in a JavaScript file that must be served at the root of the application. If it is under /static its URL will be something like /static/sw.js. It means that the scope of the ServiceWorker (ie the pages it is activated on), must be under /static. This means that it won't be enabled for our whole site (which is what we want, limiting it to /static makes it unusable on standard pages so it would be quite useless). So it must be served from something like http://wwww.example.com/sw.js.

To achieve this, I created a simple TemplateView that will serve a file named sw.js. This appears to be the best way to achieve what we set out to do

iogt/templates/base.html
Kishokanth-netizen: Agreed. But since our sw.js is inside the template folder, I do not have access to JS functions that is inside sw.js. So I had to bring the function inline inside base.html.

Kishokanth-netizen: PS- The code here however will see further changes when addressing https://github.com/unicef/iogt/issues/117

istride: Alright, I think the inlining is not a big issue, so that can stay.

I would still want the service worker to be registered as soon as the page is loaded, regardless of whether caching is required or not. It is possible to [communicate with the worker via events](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage). So I suggest that the 'download' button, when clicked, sends a message to the worker, which then sets up the caching.

istride: Response from Kish via Slack:

> so I spent a few hours yesterday trying to make a fix as per your comment in https://github.com/unicef/iogt/pull/110/files#r655496249, and make some progress as there was indeed a module called workbox-window that allowed me to 1. Load the service worker as soon as the user comes via a JS enabled browser and 2. communicate with the service worker via messages. The snag however is that the precacheAndRoute feature can only be enabled inside the install event of the serviceworker and not by listening to the messages. Ive tried many workaround for this using the documentation, but it appears that the only way to have this "cache on clicking the download button" feature is how we have done in the current implementation. We will need to register the serviceworker on the click of a button, which then begins the precaching via precacheAndRoute

PR #112, opened 2021-06-15
Questionaries

bart-merix:


dannyparsons:
@bart-merix Thanks for this. The basic functionality looks good. There are some details from the issue which need addressing or modifying. See comments below:

**Surveys**
- Instead of the "Multi-step" option there should be a "Page Break" checkbox in the Form Fields to determine a page break after that question e.g. if "Page Break" is checked on question 5 then there is a page break after question 5.
- In Form Fields, "Required" checkbox should be unchecked by default
- In Form Fields, Admin label is missing: `All fields should have an associated "Admin Name" which relates to the CSV file header when exporting survey results.`
- In Form Fields, the options for skip logic for radio buttons and drop downs is not there. I think the choices need to be their own section instead of a single text field. This would also allow for skip logic to be included. See image from v1 below for what we need on skip logic
image
- The "DESCRIPTION AT THANK YOU PAGE" section is nice. I don't think we also need "LINK AT THANK YOU PAGE" because links can go in the body of "DESCRIPTION AT THANK YOU PAGE". Unless I am missing something different?
- In Form Fields, I'm not sure what "Hidden field" is?
- Also, after submitting the survey, I just saw a blank page, it didn't show any of the thank you information.

dannyparsons:
**Polls**
- I had to create a Polls Index Page before I could create a poll. For surveys I didn't need this and I could also create a survey as a child of any section. I think this is what we want for polls as well. I'm not sure how to add a poll to a section otherwise?
- I think we also don't need the "LINK AT THANK YOU PAGE" section since links can be added in "DESCRIPTION AT THANK YOU PAGE"

istride (changes requested):
There is far too much code here for an effective review, so I would have appreciated breaking down this feature further, or building in stages. Something to bear in mind going forward. Assuming Danny's feedback is addressed, I'm inclined to just accept this PR and deal with any problems later.

However, I do have a problem with the way polls have been implemented, in that, it seems to be a carbon copy of v1, and I'm not sure it is the best way to deal with polls. Have we considered that a poll could be a form with only a single field, compared to a survey that is a form with many fields? There are behavioural differences, of course, but does that stop us from considering them both as wagtail forms? There is an opportunity to improve the way things are done compared to v1 and I'm worried that we're missing it here. I'm happy to have a call about this to go over it in more detail.

To echo Danny's feedback, the admin front-end experience is very confusing at the moment and needs to be more consistent with the way things are typically done in Wagtail.

mikowhy:
@istride ok, I understand your point of view. Let us discuss it internally and we arrange a call. Ok?

dannyparsons (approved):
Tested and working well. Further additions will be made in a new PR.

PR #115, opened 2021-06-15
Make user registration survey mandatory

### LINKED ISSUE #113: Make it impossible to bypass Custom Registration questions
The registration questions' data is very important for the countries hosting IOGT. The questions are essentially part of the registration process and no user should be able to register to the platform without completing the questionnaire. 

The custom questions need to be customisable by the editors and show up right after setting up the username password. If the user exits the website at this point, they should be 'locked out' of IoGT until they fill the survey form. They shouldn't even be able to access the public content that i accessible otherwise.

cc: @dannyparsons

ehmadzubair:
Fixes #113 

- Added Middleware to intercept requests by users who haven't filled the registration survey
- Add test cases
- Updated the language
- Removed user_passes_test decorators


iogt_users/middlewares.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        current_url = resolve(request.path_info).url_name
+        is_post_registration_survey_url = current_url == 'post_registration_survey'
+        is_registered_user = not request.user.is_anonymous
+
+        if is_registered_user and not is_post_registration_survey_url and \
+                not request.user.has_filled_registration_survey:
```
ehmadzubair, line 20: Just thinking out loud if it would be better to allow the user to at least sign out? The current implementation doesn't let them do anything.

cc: @istride @dannyparsons

ehmadzubair:
@istride

iogt_users/middlewares.py
istride: I would say that it's necessary to allow a user to sign out no matter what. I expect there to always be a link in the header for signing out/in.

dannyparsons: I agree.

ehmadzubair: @istride I've updated the PR to accommodate this

ehmadzubair:
@istride The PR is updated. This should be good to merge now.

dannyparsons:
Merging as @isedwards has reviewed the code (offline)

PR #116, opened 2021-06-15
Add RapidPro chatbot functionality

geoo89:
In order to meaningfully try this out, you will need access to a workspace on a RapidPro server (e.g. get access to IDEMS' RapidPro server) in order to set up a channel for the IoGT server to communicate with. Your workspace should also have at least one chat flow and a trigger word starting this flow (you create your own flow or import the [attached sample flow](https://github.com/unicef/iogt/files/6657851/simple_test.zip) with the trigger word 'testtrigger' into RapidPro).

In RapidPro, go to Channels (top right) -> Connect External API. You will see a form like this:
image
Fill it out as shown: The "Address" field is currently not used, so you can fill in anything. The max length should be as high as possible; on IDEMS' RapidPro server the maximum is 640, on newer versions (since Jan 7th 2021) it is 6400. The "Send URL" is the endpoint on your development IoGT server, which is http://YOUR_SERVER_URL/messaging/rapidpro_interface/ (trailing slash is important). The request body is the json content
```{"id":{{id}}, "text":{{text}}, "to":{{to}}, "to_no_plus":{{to_no_plus}}, "from":{{from}}, "from_no_plus":{{from_no_plus}}, "channel":{{channel}}, "quick_replies":{{quick_replies}}}```. The response that the RapidPro server sends upon successfully receiving a message is "All Good.". After submitting a form, you will be presented with a list of URLs which are the communication endpoints of the RapidPro server. The "Received URL" is the one of interest as this will be the one the IoGT server sends messages to.

In the IoGT website, create a ChatbotChannel entry in the corresponding DB table (e.g. using the django-admin interface).
This requires a request_url, which is the "Received URL" you got when submitting the channel form in RapidPro. The identifier is currently unused.

To create a message thread, log into the IoGT website and go to the URL /messaging/inbox. Click "New Message". You will be able to select the ChatbotChannel entry you created earlier as the bot to have a conversation with. As the message content, choose the trigger word that you defined in your RapidPro workspace. If everything was set up correctly, after a while the message thread should contain the responses from the bot; you can then continue the conversation.


geoo89:
For reference, the implementation of the RapidPro channel handler that we're using here: https://github.com/nyaruka/courier/blob/main/handlers/external/external.go

PR #130, opened 2021-06-24
Refactor Chatbot

ehmadzubair:


messaging/api/views.py
```diff
@@ -0,0 +1,39 @@
...
+        serializer = RapidProMessageSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        rapidpro_message_id = serializer.validated_data['id']
+        thread_uuid = serializer.validated_data['to']
+        text = serializer.validated_data['text']
+        quick_replies = serializer.validated_data['quick_replies']
+
+        thread = Thread.objects.get(uuid=thread_uuid)
+        # TODO: Decide how to treat each of these potential errors:
```
ehmadzubair, line 29: @geoo89 would love to discuss this on a call

messaging/api/views.py
```diff
@@ -0,0 +1,39 @@
...
+        # - Invalid thread UUID
+        # - channel UUID mismatch
+        # TODO: Extract attachments from messages.
```
ehmadzubair, line 32: @istride @dannyparsons Would want to discuss the limitations related to attachments from RapidPro

messaging/api/views.py
geoo89: Just for the record here, as we discussed, there's no need to check for the channel UUID as we're matching by thread already, and for the thread_uuid you already prepared another commit.

messaging/chat.py
```diff
@@ -0,0 +1,66 @@
...
+                'sender': sender,
+                'text': text,
+                'quick_replies': quick_replies,
+                'thread': self.thread,
+            })
+            if not created:
+                # If the message already exists, we concatenate the newly received
+                # messages with the existing message. An assumption here is that
+                # messages will always be received in the correct order. This should
+                # be confirmed with RapidPro
```
geoo89, line 29: I do not think this is guaranteed, but there's nothing we can do about it. With the message length limit set to 6400 in recent released of RapidPro, messages being split should be exceedingly rare.

geoo89 (approved):
This looks great! You already implemented the two minor fixes we discussed during our meeting. Dealing with attachments is an issue for a separate PR.

ehmadzubair:
@dannyparsons This is ready to merge. You can find instructions to run the chatbot in the README.md file

PR #142, opened 2021-06-25
Skip logic (wip)

rafalnater:


dannyparsons (approved):
Merging this so that skip logic design is in.

dannyparsons (approved):
Merging this so that skip logic view is available.

PR #171, opened 2021-06-28
Enh/enable sort for banner index, section index, and section pages

### LINKED ISSUE #153: In the CMS, pages should always display in "Sort" order
For any page in the CMS which displays pages:
- Banners index page
- Section index page (when created in #152)
- Section pages

The pages should always be displayed in the "Sort" order (which is the order they appear on the site).

This has already been done for Footer pages so the implementation can be the same.

sheralim012:
Fixes #153 

- Enabled sorting for banner index page, section index page and section page


sheralim012:

dannyparsons (approved):
Tested and works well.

PR #174, opened 2021-06-28
Add language-specific menus using flat menus

isedwards:
`Admin > Settings > Flat Menus` can now be used to create language-specific top-level menus for each required language.

When templates are rendered, the current language code will be automatically detected - this language code is then used to display the correct menu.

In order for the correct menu to be chosen, the menu's `Handle` must be set to `en_menu_live` for English, `fr_menu_live` for French, etc.

Only the overall menu requires the `Handle` to be set, the individual menu items within a menu do not require a `Handle` value.

This PR also disables (hides) Wagtail's standard "Main Menu" option from `Admin > Settings` to avoid confusion.

Resolves 137, resolves 144.

### DIFF
--- a/iogt/settings/base.py
+++ b/iogt/settings/base.py
@@ -84,6 +84,9 @@ MIDDLEWARE = [
     'external_links.middleware.RewriteExternalLinksMiddleware',
 ]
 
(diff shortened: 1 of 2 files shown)


dannyparsons (approved):
Tested and works well, thanks.

PR #177, opened 2021-06-29
Adds registration survey response in user export list

### LINKED ISSUE #141: Responses to registration survey should appear in user exports
@ehmadzubair Please take this up

sheralim012:
Fixes #141 

- Added registration survey response in users list export
- Only the latest submission will be included in users list export

### DIFF
--- a/iogt_users/wagtail_hooks.py
+++ b/iogt_users/wagtail_hooks.py
@@ -1,6 +1,7 @@
 from django.contrib.auth import get_user_model
 from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register, ModelAdminGroup
 
+from home.models import SiteSettings
 from iogt_users.filters import GroupsFilter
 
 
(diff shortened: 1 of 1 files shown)


sheralim012:

dannyparsons:
@sheralim012 Can we have the responses to the custom questions in a separate column for each question instead of a list in a single column?

sheralim012:
@dannyparsons user response for the survey is stored as JSON object in text field like the following pic
Screenshot 2021-06-29 at 4 45 40 PM

to include questions and responses as individual columns we need to override the view which generates the file for export and also as we change and add new columns data will look like a sparse table too.

Let me know if you want me to go ahead and do this.

dannyparsons:
@sheralim012 When I go to the Form Data page and export the data from the registration survey on it's own it's in a nice format with a column for each question. Can we reuse this functionality here too?

dannyparsons:
This needs improvement for better formatting, but is an improvement on the existing functionality.

PR #179, opened 2021-06-29
Embed Polls and Surveys in Article Pages

### LINKED ISSUE #154: Surveys, Polls and Quizzes can be embedded within an article
The body field in articles should have an additional block or blocks for embedding a survey, poll or quiz.

If the survey or quiz has any page breaks these are ignored and it will display the whole survey/quiz.

When submit is clicked the thank you page also displays embedded. On a feature phone, this will require a page refresh.

ehmadzubair:
Fixes #154 

 - Added embedded_questionnaire.html template
- Added EmbeddedQuestionnaireChooserBlock
- Embedded Surveys and Polls in Articles

## Testing Instructions
Add embedded Surveys and Polls in the body field of an Article


dannyparsons (approved):
Tested and works well, thanks.

PR #180, opened 2021-06-29
Update page choose in article body to button with text

### LINKED ISSUE #164: Page link block in body of article should have editable link text
When inserting a page link as a block in the body of an article the link text should be editable - default as the page name.

The field should also be renamed to "Page Button" as it will be displayed as a button.

sheralim012:
Fixes #164 

- Updated page chooser to Page Button struct block in wagtail admin
- Page Button struct block will contain 2 things
  - User facing text for button
  - Page chooser option
- If user-facing text is empty then page title will be used as user-facing text for button


sheralim012:

home/blocks.py
```diff
@@ -52,3 +52,11 @@ class SocialMediaShareButtonBlock(blocks.StructBlock):
 
     class Meta:
         icon = 'site'
+
+
+class PageButtonBlock(blocks.StructBlock):
+    text = blocks.CharBlock(required=False, max_length=255)
+    page = blocks.PageChooserBlock()
```
dannyparsons, line 59: Can you just change the order of these so page is first?

dannyparsons (commented):
Works well nice. Just one trivial comment below.

Also, if it's easy to add to the "Text" field something like "(optional - leave blank for page name)" that would be good.

home/blocks.py
dannyparsons:
```suggestion
    page = blocks.PageChooserBlock()
    text = blocks.CharBlock(required=False, max_length=255)
```

PR #188, opened 2021-06-30
Feat/quizzes

### LINKED ISSUE #38: As a CMS Editor, I can configure quizzes for front-end users to test their knowledge
Formatted similarly to surveys but with the option to identify correct responses for positive number, number, date, drop-down, and radio button fields. The quiz feature comes with an automated feedback loop for users sharing their score. In the CMS, editors can craft messages for when the user answers incorrectly.

sheralim012:
Fixes #38 

- Added quizzes
- Currently only simple field equality check is implemented to check correctness of answers of user (added a TODO)


sheralim012:
@dannyparsons kindly check and let me know if any change is necessary

dannyparsons (approved):

PR #190, opened 2021-06-30
Cleaned up migrations

ehmadzubair:


ehmadzubair:
Updated this with the latest develop branch

PR #205, opened 2021-07-01
Adds child quizzes in section page

sheralim012:
- Added child quizzes in section


sheralim012:
@dannyparsons added child quizzes in section as discussed before.

PR #214, opened 2021-07-01
Adds wagtail svg support

sheralim012:
- Added wagtailsvg package
- Changed flat menu item icon to svg
- Added svg icon in section and article (footer is basically article, so it will automatically get it)


sheralim012:

PR #220, opened 2021-07-02
Quiz can be multi-step now, and fixes skip logic issue

sheralim012:
- Enabled multi-step for quizzes
- Fixed skip logic


sheralim012:
@dannyparsons could you please check and let me know.

PR #239, opened 2021-07-06
Register urls for precache

sheralim012:
- Uses sitemap API to fetch URLs to use in precache
- All Pages (homepage, sections, articles, footers, polls, surveys, quizzes) will be cached


sheralim012:

PR #253, opened 2021-07-07
Sticky footer added

Kishokanth-netizen:


geoo89:
Merged all the commits except for the last one (84d110f) which doesn't have functionality, only trying to restore files that were altered involuntarily.

PR #260, opened 2021-07-08
Adds mgmt cmd to add index pages

sheralim012:
- Added management command to run during migrations to add index pages of banners, footers and sections


ehmadzubair:
Looks good.

cc: @geoo89

PR #262, opened 2021-07-08
Cleaned the SCSS

Kishokanth-netizen:


geoo89:
The footer now only extends across half the screen width for me, and a non-clickable "Sections" text appeared as well. Could you have a look at that? See attached.
image

PR #265, opened 2021-07-11
Fix misplaced sections placement

ehmadzubair:
Some sections have been misplaced as part of the initial data entry. This command will fix their placement to be direct children of the SectionIndexPage.


ehmadzubair:
@volloholic

PR #267, opened 2021-07-12
Added banners and footers content migration

sheralim012:
- Added banners in content migration (without page links)
- Added footers in content migration
- Simplify wagtail page tree fix
- Simplify index pages creation
- Made site from v1 default site

### DIFF
--- a/iogt_content_migration/README.md
+++ b/iogt_content_migration/README.md
@@ -5,6 +5,9 @@
 ### Prerequisites
 
 - v1 database backup
+```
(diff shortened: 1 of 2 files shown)


sheralim012:
@volloholic @geoo89 kindly check and let me know.

ehmadzubair:
👍🏼

sheralim012:
also @volloholic, it would be helpful if we can get the dump of v1 db and test it on the local machine before running on prod

Issue #114, opened 2021-06-15
CMS Editors can set Site Manifest Options


dannyparsons:
The main fields of the site manifest https://github.com/mikowhy/iogt/blob/feat/IU-11-offline-access-implementation/iogt/static/manifest/manifest.webmanifest for offline access should be editable by CMS users. This could be a specific page in the CMS.

geoo89:
Resolved by PR #269

PR #269, opened 2021-07-12
Feat/iu 28 admin panel for manifest (backend part)

mikowhy:
new solution for manifest admin panel → https://github.com/unicef/iogt/issues/114
---
## steps to be taken to create webmanifest

- add manifest → http://localhost:8000/admin/home/manifestsettings/ (if not added, 404 will appear)
- added manifest can be found here http://localhost:8000/manifest.webmanifest
- it is seen under network tab 

image

**details:**
- 1 manifest per language
- icons: png files + dimensions validator


sheralim012:
LGTM 👍🏽

Issue #156, opened 2021-06-26
Features to add to Surveys


dannyparsons:
Front-end:
- [x] Support for URL input fields. Currently, they don't appear
- [x] When input type "checkbox" is set as "required" in the backend, when submitting, the error message "This field is required." is displayed regardless of whether it is checked or not, making it impossible to complete the survey.
- [x] Support for "(single) Checkbox" fields. They currently crash the page.
- [x] Support for "multiple select" fields. The selected options currently don't display properly.
- [x] Bugfix: If there are multiple "dropdown"/"multiple select" fields on a single page, clicking on a lower one will show the options from the first one.
- [x] Bugfix: If there are multiple "checkboxes" fields on a single page, clicking on an option (i.e. the text, not the checkbox itself) from a lower one will check/uncheck the corresponding option for the first one.
- [x] Bugfix: required text fields are currently marked with the tag "optional"
- [x] Bugfix: When receiving a validation error (e.g. empty required field), and being redirected to the same page of the questionnaire, the numbering of questions is wrong (e.g. Question 2 of 1)
- On the first page of a survey/quiz/poll, the "Back" button takes you to the home page rather than back to the article/section you came from. Now #359
- [x] When input type "checkbox" is set as "required" in the backend, the info text is "CHECK ALL THAT APPLY". 

Backend:
- [x] Bugfix: In the admin interface, "Answer options" is a required field (should have at least one entry, otherwise a validation error occurs when trying to save). However, when saving a question with text/number/etc input, "Answer options" for such fields is erased. I believe it shouldn't be a required field for these input types, or maybe not be available at all? Also see #301 
- Implement temp fix for #315 by disabling answer options for input type "checkbox"
- [x] Page breaks need to functional when surveys are published
- [x] Skip logic needs to be functional when surveys are published
- [x] Support _Positive Number_ in addition to _Number_ input type. Positive Number has 0 as the minimum.
- [x] Support for anonymous submission (if specified in the backend).

ehmadzubair:
I'm starting work on this ticket

geoo89:
Crash log for "checkbox" input type as of commit 948d8a2:
This is for a survey with a single checkbox.

Environment:


Request Method: GET
Request URL: http://127.0.0.1:8000/en/sections/youth/the-survey/

Django Version: 3.1.11
Python Version: 3.6.9
Installed Applications:
['home',
 'search',
 'iogt_users',
 'comments',
 'iogt_content_migration',
 'questionnaires',
[...]

PR #271, opened 2021-07-12
Updated description text for the question

Fourmeta:


geoo89 (changes requested):
I also noticed that if you have multiple Dropdown (select) menus on the same page in a survey/quiz/poll, and I click on any but the first one, the dropdown list from the first one will be displayed.

In addition, it seems like the input type "Checkbox", "Multiple select" and text (different flavors) fields currently don't work (checkbox crashes the page, text field doesn't show up in the quiz view; they do show up for surveys). This doesn't have to be addressed in this PR though.

Also refer to Issue #156

PR #278, opened 2021-07-13
Updated styles for the textarea

Fourmeta:


geoo89:
I resume this PR is in response to the drop down menus not working on feature phones?
Now the dropdown menus look basic on desktop as well, which clashes with the rest of the designs.
Do you think it is possible to use the previous version on desktop browsers while displaying the basic dropdown on feature phones?

PR #282, opened 2021-07-13
Updated styles for articles

Fourmeta:


geoo89:
On its own, this PR (in particular commit aba3421 I think) seems to break dropdown menus.
I assume this should go together with PR #277?

PR #284, opened 2021-07-14
Fixed bug with checkboxes

Fourmeta:


geoo89:
Just to make sure we're on the same page:
- I should merge PR #278 before testing this one
- I can close PR #281 and PR #282  because this PR contains all the changes from these
is that correct?

Let me know once you're done with changes, so I can have a closer look.

geoo89:
I updated this dev server to include your changes, and I set up a couple of test questionnaires: http://iogt-django-dev.idems.international/en/sections/questionnaire-testing/

Later today I'll update the issues to check off the things that work now and add some other problems I noticed, and merge this PR.

PR #313, opened 2021-07-18
Add missing localization to user facing sites

geoo89:
I'm not sure localizing the URLs is the way to go, so please check and we can undo the changes to iogt/url.py as needed. I've noticed some issues already, e.g. http://localhost:8000/fr/accounts/login/ redirects to http://localhost:8000/en/users/profile/ (i.e. the language gets changed), and the "Home" buttons that were added redirect to '/', i.e. the English version of the site.

There are a couple of instances of {% translate err %} which I left as is, as I believe these contain a variety of error messages. I haven't checked where these error messages are indexed though (so that i18n can pick them up). It seems like django-forms/wagtail is already picking them up and translating (e.g. Enter a valid date).

These following two files contain specific articles and I believe they are not in use (so I didn't localize). Can someone confirm, so we can delete them?
- home/templates/home/tags/polls.html
- home/templates/home/tags/questionnaire.html

iogt/templates/base.html has some untranslated code in a JS tag, not sure what to do about it.

There are a couple of message when putting invalid input in text fields, such as 
- Please include an '@' in the email address. (e-mail field)
- Please enter a URL (URL field)
- Please enter a number (number field)
I believe this is actually the native browser validation, and thus will be localized according the user's language that is set in the browser, which I assume it good enough (and I don't think we could change them, at best remove them, see https://github.com/logaretm/vee-validate/issues/1595 )

Some localization strings already seem to provided by wagtail, e.g.:
- Enter a valid date. (error when submitting invalid date in form)
- Your comment (pre-filled text in comment field)
- Back (which I'm overwriting in this PR)


home/blocks.py
```diff
@@ -13,25 +14,40 @@ def render_basic(self, value, context=None):
         if not value:
             return ''
 
+        video_not_supported_text = _("Your browser does not support the video tag.")
+        audio_not_supported_text = _("Your browser does not support the audio element.")
+        # Translators: Part of this message (between %(start_link)s and %(end_link)s ) is a clickable download link
+        download_video_text = _('If you cannot view the above video, perhaps would'
```
ehmadzubair, line 20: It's normally better to use f strings for creating templated strings. It's more readable

home/blocks.py
```diff
@@ -13,25 +14,40 @@ def render_basic(self, value, context=None):
...
+                ' you like to %(start_link)s download it? %(end_link)s.') % {
+                        'start_link': '<a href={1} download>',
+                        'end_link': '</a>'
+                }
```
ehmadzubair, line 24: I don't know if this would translate well (and don't know the best practice to do this yet). Have you tested this?

geoo89: Yes, this works (as long as we use gettext rather than gettext_lazy, see definition of _ at the top).

home/blocks.py
geoo89: The last paragraph from [here](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#standard-translation) states:

"Since string extraction is done by the xgettext command, only syntaxes supported by gettext are supported by Django. In particular, Python f-strings are not yet supported by xgettext, and JavaScript template strings need gettext 0.21+."

Thus I will leave this as it is.

ehmadzubair (commented):
I just realized that this review wasn't submitted

PR #314, opened 2021-07-19
Added desktop layout

ehmadzubair:
This PR includes:
- Complete desktop layout with 1,2,3 column views on different sizes
- Hide/Show relevant buttons on specific layouts
- Refactored `header.html` to unify the header code

Limitations:
- Most of the inner elements have their own padding and margins which are causing some different in the layout. We need to do a complete pass on the inner elements too
- This hasn't been tested on production data
- The language picker needs to have a css only based solution (doing that in a ticket)


ehmadzubair:
Fixes #333 

@geoo89 this is ready to merge

PR #326, opened 2021-07-20
Added direct display to Questionnaires

ehmadzubair:
Fixes Direct Display from #6 

As part of this ticket, the section screens will now respect the direct_display flag. If that is set to `True`, a form will be rendered

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -209,10 +209,17 @@ class Section(Page, PageUtilsMixin):
             self.featured_content.filter(content__live=True)
         ]
         context['sub_sections'] = self.get_children().live().type(Section)
+
         context['articles'] = self.get_children().live().type(Article)
-        context['surveys'] = self.get_children().live().type(Survey)
-        context['polls'] = self.get_children().live().type(Poll)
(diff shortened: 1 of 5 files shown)


home/models.py
```diff
@@ -209,10 +209,17 @@ def get_context(self, request):
             self.featured_content.filter(content__live=True)
         ]
         context['sub_sections'] = self.get_children().live().type(Section)
+
         context['articles'] = self.get_children().live().type(Article)
-        context['surveys'] = self.get_children().live().type(Survey)
-        context['polls'] = self.get_children().live().type(Poll)
-        context['quizzes'] = self.get_children().live().type(Quiz)
+
+        survey_page_ids = self.get_children().live().type(Survey).values_list('id', flat=True)
```
sheralim012, line 215: shouldn't we filter out polls, surveys, and quizzes which are multi-step as multi-step forms don't work in direct display?

ehmadzubair: Good point. I think the better way to do this would be to add validation in the Admin page. To not allow direct display for multistep forms. Will make a ticket

ehmadzubair: #329

Issue #317, opened 2021-07-19
Design for external link referral page


geoo89:
The page exists and is functional, but the text and design need to be finalized.

Current state:
![image](https://user-images.githubusercontent.com/678199/126184964-c73dd2fa-ac57-45f8-913a-2a14725e7841.png)

geoo89:
This could still do with some improvement.

ChrisMarsh82:
This will now be customizable.

geoo89:
Each country office should be able to customize this page, and create one for each of their locales.

At the minimum, there should be 3 text fields (defaults for English in quotes):
Main text: "You are leaving the Internet of Good Things to visit an external website and standard data charges by your network provider might apply"
Continue button text: "Continue to external site"
Back button text: "Back"

Ideally though, rather than only configuring the main text, it should be possible to design this page like an article, with the Continue/Back button text as special fields to be entered, which will be rendered as buttons below the article.

geoo89:
Add it in the site settings like the post registration survey.

PR #340, opened 2021-07-25
Bare but sensible transition page

geoo89:
image

Temp fix for #317


geoo89:
We don't have locale specific logos, so use text instead of embedding the logo.

PR #344, opened 2021-07-26
skip content for locales other than eng

sheralim012:


geoo89:
Thanks, this is perfect!

PR #363, opened 2021-07-27
move content to correct locales during content migration

sheralim012:


geoo89:
Thanks, this seems to be working really well. I'll set up VMs tonight with migrated content so people can test more thoroughly.

Only minor thing I've noticed is that the few banners that were there seems to be missing their links, but that might be easy enough to fix manually.

PR #367, opened 2021-07-28
adds missing migration

sheralim012:


sheralim012:
@geoo89 this adds missing migrations, check and let me know

PR #368, opened 2021-07-28
fixes polls landing page, removes multiselect

### LINKED ISSUE #339: Temporarily remove 'checkbox'/'multiple select' from polls backend
While they work for surveys, the 'checkbox' input fields crash the page for polls.
As they are of limited use for polls, disable the 'checkbox' option temporarily until this is fixed.

Also disable "Multiple select" from surveys, polls and quizzes for now. They are not part of the formal requirement, and need some polishing (or checkboxes might be sufficient).

sheralim012:
Fixes #339 

- Fixes polls rendering issue
- Removes multiselect field type from polls, surveys, and quizzes


geoo89:
It looks like you removed dropdown from polls, which we want to keep. (Instead, you can remove checkbox.)

PR #374, opened 2021-07-29
separates poll results for checkboxes

### LINKED ISSUE #360: Poll results for input type checkboxes
For input type checkboxes (and multiple select), poll results are displayed as lists, rather than aggregating the elements of the list, see example where the answer options were 1, 2 and 3
image

sheralim012:
Fixes #360 

- Polls multiselect fields like checkboxes results will be calculated based on each individual choice
- Added missing check for poll submission so results are only calculated after poll submission


ehmadzubair:
👍🏼

PR #377, opened 2021-07-29
adds validation for page break with multi step

### LINKED ISSUE #354: Page breaks crash quizzes/surveys
If multi-step is enabled, each question is on a new page, and everything works fine (manual page breaks are ignored).

If multi-step is disabled, and a manual page break is inserted somewhere in a quiz/survey, the page crashes.

Should be fixed or disabled (I don't see page breaks in the requirements, but I'm not sure).

sheralim012:
Fixes #354 

- If page break is enabled for any question without multi-step enabled, a validation error will be raised

### DIFF
--- a/questionnaires/forms.py
+++ b/questionnaires/forms.py
@@ -36,6 +36,11 @@ class SurveyForm(WagtailAdminPageForm):
             self._clean_errors = {}
             if form.is_valid():
                 data = form.cleaned_data
+                if self.data.get('multi_step', 'off') == 'off' and data['page_break']:
+                    self.add_form_field_error(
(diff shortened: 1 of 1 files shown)


ehmadzubair:
👍🏼

geoo89:
I wonder if this is sufficient. What happens if multiple questions on the same page have skip logic, which one is chosen?
I would propose in that case to let the last question on the page determine the result of the skip logic, and put a note about that in the help text.

Also I noticed, currently for quizzes, the page breaks are ignored, and every question is served on a separate page, if multi-step is enabled. This should be fixed too.

PR #410, opened 2021-08-04
migrate polls and surveys to appropriate section

sheralim012:


sheralim012:
@geoo89 check and let me know

PR #398, opened 2021-08-02
Added reply and like buttons

ehmadzubair:
Fixes #385

PR #435, opened 2021-08-10
fixed report button positioning

ehmadzubair:


geoo89:
Is this obsolete? The location of this button has already been changed in #398 it seems.

ehmadzubair:
Good catch @geoo89

PR #438, opened 2021-08-10
display section icon for flat menu item

### LINKED ISSUE #148: Flat Menu item can automatically display section icon for a section page link
If a menu item in a Flat Menu is linked to a Section page, then it should not be necessary to choose the icon for the menu item and it should use the section icon when the item icon is blank. This can be overridden by specifying an icon for the menu item.
The icon field should have the hint text:

> If Page link is a section page and icon is blank then the section icon will be used. Specify an icon here to override this.

sheralim012:
fixes #148 

- if flat menu item icon is available then display it
- if flat menu item icon is not available and item is section page then display section icon 
- add help text for flat menu item icon field


ehmadzubair:
👍🏼

Issue #421, opened 2021-08-05
Anchors are removed from URLs when using External Links


cbunicef:
Created an external link to 

http://central.iogt.site/en/sections/your-health/mental-health/keeping-mentally-healthy-self-care-tips/#its-going-to-be-alright

After the "Leaving site" warning, the destination URL is

http://central.iogt.site/en/sections/your-health/mental-health/keeping-mentally-healthy-self-care-tips/

geoo89:
Fixed by #444

Issue #420, opened 2021-08-05
"Leaving site" warning shows when user follows an "external link" which directs to the IoGT site


cbunicef:
Steps to reproduce

1) create External Link in article that links to a page within the IoGT site
2) click the link in the article
3) "Leaving site" warning appears

Desired behavior

1) create External Link in article that links to a page within the IoGT site
2) click the link in the article
3) No "Leaving site" warning when link destination is the same ... domain subdomain? Not sure about this one

It's useful to have External Links with destination inside the site in order to send users to an anchor in another article.

ehmadzubair:
You can add internal links using the "Internal Link" tab from the link. The external link tab is designed to show the intermediate screen. 

![Screenshot 2021-08-10 at 3 04 18 PM](https://user-images.githubusercontent.com/10597112/128848772-675ccc27-0e12-4911-bf50-5ff0a68af9e3.png)

Does this work?

geoo89:
Internal links don't support anchors, but #444 fixes anchors being dropped from external links and skips the redirect page for external links if they go back to the IoGT site, so external links can now be used for this purpose.

One caveat here is that using external links that redirect to the IoGT site will require being change if an article is transferred to a different IoGT site (e.g. a country site).

cbunicef:
Should this be closed?

PR #444, opened 2021-08-10
Fixed internal and external urls to support fragments

ehmadzubair:


geoo89:
Fixes #420
Fixes #421

PR #453, opened 2021-08-11
Fixed heading size

### LINKED ISSUE #390: Header content block text is not larger than paragraph content block text
Input

image

Live article

image

ehmadzubair:
Fixes #390

### DIFF
--- a/home/static/css/global/global.css
+++ b/home/static/css/global/global.css
@@ -374,6 +374,10 @@
   }
 }
 
+.content .footer-main {
+  width: 100%;
+  max-width: inherit;
+}
 
 .search__form {
   background: #F7F7F9;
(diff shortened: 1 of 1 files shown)


geoo89:
It looks like this for me now:
image

So the heading and the title are the same size now. Could you choose a size that is in between the title and the paragraph size?

azahid89:
@geoo89 The title is basically an H1 tag (in the figma design). H2, H3 and H4  are available in the paragraph field. Now we can either increase the Title size or decrease the Heading size. Please note this would mean the heading will not be a H1 tag but a custom div.

geoo89:
I just tried inserting a H2 into a paragraph, and it appears that it is also the same size as the title:
image

From what I see in the page source, the title is H1, the heading is a custom div, and a H2 in a paragraph is a H2, so at least in principle there shouldn't be an issue giving each a different size. But maybe all of them being the same size is not a problem. Let me ask at tomorrow's meeting and I'll get back to you.

geoo89:
This is good for now. I created #464 to check back whether all headings having the same size is want we want, design-wise.

PR #454, opened 2021-08-11
update prompt text size

ehmadzubair:


geoo89:
It looks like this for me now:
image

I can't see any change in the prompt text size (highlighted in red).

ehmadzubair:
@geoo89 I have tested this on my local. Is this what you're looking for? If yes, can you try force refreshing?

Screenshot 2021-08-11 at 3 57 23 PM

geoo89:
Ah, it appears that for embedded questionnaires the text has increased, but not for standalone questionnaires.

ehmadzubair:
@geoo89 can you test this now?

PR #455, opened 2021-08-11
fixes bug of choices spliting due to comma

### LINKED ISSUE #389: Commas in quiz answer option text break option into multiple pieces
image

image

in this quiz: http://beta.iogt.site/en/sections/your-health/mental-health/mental-health-trivia/

sheralim012:
fixes #389 

- fixes bug of choices splitting due to comma in a single choice

Screenshot 2021-08-11 at 11 22 35 AM

### DIFF
--- a/questionnaires/forms.py
+++ b/questionnaires/forms.py
@@ -91,7 +91,7 @@ class SurveyForm(WagtailAdminPageForm):
             if field_type in VALID_SKIP_SELECTORS:
                 choices_values = []
                 for skip_logic in form.instance.skip_logic:
-                    choices_values.append(skip_logic.value['choice'])
+                    choices_values.append(skip_logic.value['choice'].replace(',', u'\u201A'))
                 form.instance.choices = ",".join(choices_values)
 
             if field_type not in VALID_SKIP_SELECTORS:
@@ -212,7 +212,7 @@ class QuizForm(WagtailAdminPageForm):
             if field_type in VALID_SKIP_SELECTORS:
                 choices_values = []
                 for skip_logic in form.instance.skip_logic:
-                    choices_values.append(skip_logic.value['choice'])
+                    choices_values.append(skip_logic.value['choice'].replace(',', u'\u201A'))
                 form.instance.choices = ",".join(choices_values)
 
             if field_type not in VALID_SKIP_SELECTORS:


ehmadzubair:
Clever :) 👍🏼

geoo89:
This solves the specific issue, but this hack is an indication of a deeper problem in the code.
For example, if the phrase with the comma is the correct answer, we cannot list it as the correct answer for the quiz.
Similarly, polls cannot have commas in their options.

Are any of these comma-separated strings stored in the database, or are they encoded differently (json, StreamField, etc)? If they are encoded as comma separated strings, this is probably something to be fixed asap.

We should either ensure that everything that is a list is also treated as a list, or at the very least that there's a way to escape commas that is consistently used throughout.

sheralim012:
@geoo89 applying the same solution to the correct answer field will resolve the issue.

questionnaires/forms.py
```diff
@@ -86,6 +86,7 @@ def clean(self):
     def save(self, commit):
         # Tidy up the skip logic when field cant have skip logic
         for form in self.formsets[self.form_field_name]:
+            form.instance.correct_answer = form.instance.correct_answer.replace(',', u'\u201A')
```
geoo89, line 89: Surveys don't have a correct answer field, so this line should be removed.

geoo89 (commented):
I don't fully understand how this works, in particular how it detects in the correct_answer field which of the commas are delimiters and which are part of a choice, but it seems to work.

However, if multiple answers are correct, even if I get all of them right (and they all show up in green), my results will say 0/1 though it should say 1/1. This should be fixed, as it was working before.

questionnaires/forms.py
sheralim012: dont

sheralim012:
> I don't fully understand how this works, in particular how it detects in the correct_answer field which of the commas are delimiters and which are part of a choice, but it seems to work.
> 
> However, if multiple answers are correct, even if I get all of them right (and they all show up in green), my results will say 0/1 though it should say 1/1. This should be fixed, as it was working before.

so I think we shouldn't use `,` comma as separator but some other character like `|` pipe

ehmadzubair:
Fixes #465

ehmadzubair:
👍🏼

sheralim012:
fixes migrations so before pulling new commits run `python manage.py migrate 0010`, then pull and then run `python manage.py migrate`

PR #463, opened 2021-08-12
Fixed the Reply screen

### LINKED ISSUE #443: Reply to comment page
image

- [ ] Remove the strange icon
- [ ] Use the same date format as in the main comment thread
- [ ] Display the username of the poster rather than firstname/lastname

ehmadzubair:
Fixes #443 

Screenshot:
Screenshot 2021-08-12 at 5 22 40 PM

### DIFF
--- a/comments/templates/django_comments_xtd/reply.html
+++ b/comments/templates/django_comments_xtd/reply.html
@@ -2,6 +2,7 @@
 {% load i18n %}
 {% load comments %}
 {% load comments_xtd %}
+{% load humanize %}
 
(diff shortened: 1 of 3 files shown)


ehmadzubair:
Although I think if we just copy the original comment here, that would be better UI experience

geoo89:
What do you mean? It looks good to me as is.

ehmadzubair:
I mean, if we can copy over the UI from the original comment, that would be ideal. If it looks good as-is, you can just merge it

PR #478, opened 2021-08-20
handles multistep child and embedded questionnaire

### LINKED ISSUE #329: Add validation to Questionnaires for Direct Display
Direct display support is added for Questionnaires in #326. While all other questionnaires work as expected, multistep forms in particular break when they are rendered elsewhere.

We should have a validation step that prohibits editors to enable direct display in multi-step forms

sheralim012:
fixes #329 

- handles multistep questionnaire as child form
- handles multistep questionnaire as embedded form


ehmadzubair:
👍🏼

Issue #465, opened 2021-08-12
Investigate storage format of poll/survey/quiz choices


geoo89:
How are the choices of questions with multiple answer options stored for polls, surveys and quizzes?
How are the correct answers stored for quizzes?

In the input forms, these are (partially?) processed via comma separated strings. This is an issue if the choices contain commas themselves. While we should find a robust solution around this, the first step is to make sure no such comma separated strings are stored in the databases, and if so, change to a format that does not have this issue asap.

This is related to #389 and the discussion in #455

cbunicef:
In data from V1 I've seen the standard comma be replaced by a unicode character looking the same but presumably not being recognized as a comma to separate values.

ehmadzubair:
Use a vertical pipe character to split options `|`

Issue #358, opened 2021-07-27
Skip logic design questions


geoo89:
From #6: "Positive number, radio buttons & drop down should provide options for skip logic within the survey configuration."

Is this accurate?
- I assume skip logic for positive numbers should be by equality?
- What is the rationale for choosing "positive number", but not "number" or the other text input fields? Should skip logic be disabled for all other field types?
- In particular, can we disable skip logic for input type "checkbox" entirely?

I assume skip-logic should only be available when multi-step is selected, i.e. one question per page? (It wouldn't make sense to have multiple questions with skip logic on one page, though one could in principle validate the only one question per page has a skip logic).

Is skip-logic a requirement for quizzes as well? If yes, how does the scoring and the results page work?

sheralim012:
@geoo89 any update or finalized requirements for this?

ChrisMarsh82:
a positive number, number, anything with a single item should support skip logic.
do not implement for quizzes.

PR #493, opened 2021-08-24
Enh/checkbox choices

### LINKED ISSUE #315: Bugfix/feature: The "Answer options" is required to have exactly two choices (true, false) for input type "checkbox". Is this for the purpose of skip logic? When saving, the choice labels are erased (though the two choices remain in the list at least). Should it be possible to omit the "Answer options" field for checkboxes if not planning to use skip logic?

sheralim012:
fixes #315 

- autofill checkbox for quiz and survey
- update error message for checkbox
- update helptext for checkbox


geoo89:
I assume this PR subsumes PRs #455 and #491, and so should fix Issues #315, #465, #389 and #358.

Before I can merge this, a few things need to be fixed:
- [ ] When changing the Field type, all choice labels are erased (or replaced with false/true in the case of checkbox). Desired behavior: If something is already there, it shouldn't be touched.
- [ ] Checkbox in survey with skip logic gives validation error: Skipping to end not allowed for checkbox and checkboxes./Skipping to question not allowed for checkbox and checkboxes.
- [ ] Hint for polls should be updated from comma to pipe: "Comma separated list of choices. Only applicable in checkboxes, radio and dropdown."
- [ ] When putting skip logic for a text field, if the user enters any value that is not covered by the skip logic crashes the page with "'answer' is not in list"
- [ ] Skip logic choices aren't saved for input type 'number'
- [ ] Conflicting migrations caused by merge commit 4b880bf

sheralim012:
@geoo89 this PR builds on top of those PRs #455 and #491 but shouldn't be merged before those.
these are interdependent and should be merged in succession starting from #455, #491, and then this one.
Issues related to those PR will be addressed in those PR, so review could be easy.

sheralim012:
> * [ ]  Hint for polls should be updated from comma to pipe: "Comma separated list of choices. Only applicable in checkboxes, radio and dropdown."

help text for poll choices field updated in #455 and as polls only have these 3 field types "checkboxes, radio and dropdown" so I think there is no need to mention them in help text too. Updated help text is just "Pipe (|) separated list of choices."

sheralim012:
> * [ ]  Checkbox in survey with skip logic gives validation error: Skipping to end not allowed for checkbox and checkboxes./Skipping to question not allowed for checkbox and checkboxes.

addressed in #491

sheralim012:
> * [ ]  Skip logic choices aren't saved for input type 'number'

addressed in #491

sheralim012:
> * [ ]  When putting skip logic for a text field, if the user enters any value that is not covered by the skip logic crashes the page with "'answer' is not in list"

addressed in #491

sheralim012:
> * [ ]  When changing the Field type, all choice labels are erased (or replaced with false/true in the case of checkbox). Desired behavior: If something is already there, it shouldn't be touched.

addressed in #493

PR #505, opened 2021-08-28
refactor fields rendering, default value rendering and retention of entered value on error

### LINKED ISSUE #486: Poll/Survey/Quiz: User input erased upon validation error
When one user input is invalid upon submitting and the page is reloaded to display the validation error, the entire user input for all input fields gets erased.

### LINKED ISSUE #487: Labels for poll questions
For polls/survey/quizzes, each question has a question label ("label" form field in the admin view).

For polls, the "label" of the form field is ignored (except for input type checkbox, where it is displayed next to the checkbox).
We should be consistent and display questions the same way across polls/survey/quizzes, so in for polls the question label should be displayed in front of the question (except for checkbox where it is already displayed next to the checkbox, which is good).

### LINKED ISSUE #498: Don't display section titles with each poll.
For polls within sections (not direct display), the section title is displayed:
Before:
image
Now:
image

It appears that this was introduced in 52ae04f93814b3150e597

### LINKED ISSUE #499: Direct display polls crash
'PollFormField' object has no attribute 'has_skipping'

See: http://iogt-django-dev.idems.international/fr/sections/questionnaire-testing/subsection/

If possible, could you factor out the commonalities between normal and embedded questionnaires as part of this ticket?
We're running into a lot of obscure issues like this, and at this point I think the testing overload caused by this is larger than the time it'd take to do a bit of refactoring.

### LINKED ISSUE #503: Refactor help text for (embedded) polls/survey/quiz
PR #495 allowed to display custom help texts for questionnaire input fields.

However, the help text doesn't display if the questionnaire is embedded.

This is because there are two more instances of this logic in home/templates/blocks/embedded_questionnaire.html which wasn't changed. Rather than just adding it there as well, could you please factor this logic out?
(I believe the logic for the embedded setting is a bit broken right now anyway, e.g. displays "select one" for checkboxes in quizzes.)

(This is from a response to PR #495 which I accidentally merged in.)

### LINKED ISSUE #504: Dynamically update 'answer options' label in survey depending on input type
image

For checkboxes, radio and dropdown the labeling "Answer Options" and "Choice" is ok.

For the other input types however, the meaning of these fields is different and pertains to skip logic.
Thus when the user selects an input type that is NOT checkboxes, radio or dropdown, dynamically update as follows:
- rename "Answer Options" to "Skip Logic options"
- rename "Choice" to "Skip Value"

### LINKED ISSUE #508: Add validation to prevent pipe symbol in answer choices
The choice field must not contain pipes (|), as choices with pipes are split into lists with multiple elements:
image

Add a validation that ensures this.

sheralim012:
fixes #486, 
fixes #498, 
fixes #487, 
fixes #503, 
fixes #499, 
fixes #504, 
fixes #508

- refactor and simplify fields rendering
- fixes date and datetime field rendering
- if default value is given for field then autofill it
  - update help text for default value to use pipe (|) for checkboxes
- retain entered values on error
- fix section name display for polls start button
- unify poll, survey, and quiz UI for label and question count and their CSS
- refactor help text for field
- fix poll direct display crash
- dynamically change answer option label
- add validation for pipe symbol in choice of skip logic


geoo89:
I managed to crash a survey with skip logic. Steps to reproduce:

Q1: Question with checkbox, checking the checkbox goes to Q3
Q2: anything
Q3: Dropdown: Option 1 goes to end, Option 2 goes to next default question.
Q4: anything

- Go through the survey, check the checkbox in Q1. Pick option 2 in Q3. Q3 will be repeated (this is a bug too). Pick anything for Q3 and Q4. Upon submission of Q4 the page crashes: "list index out of range" (related to the pagination)
- Regardless of your choices, the question counter never seems to be beyond "question 2 out of 4"

geoo89:
> dynamically change answer option label

This doesn't appear to be working?

Everything else seems to be working great though, excellent!

sheralim012:
> > dynamically change answer option label
> 
> This doesn't appear to be working?

could you try after clearing the browser cache, or try in incognito?

geoo89:
Thanks, I managed to get it to work, not sure what the issue was.

This only leaves the skip logic crash from the previous comment.

sheralim012:
> I managed to crash a survey with skip logic. Steps to reproduce:
> 
> Q1: Question with checkbox, checking the checkbox goes to Q3
> Q2: anything
> Q3: Dropdown: Option 1 goes to end, Option 2 goes to next default question.
> Q4: anything
> 
> * Go through the survey, check the checkbox in Q1. Pick option 2 in Q3. Q3 will be repeated (this is a bug too). Pick anything for Q3 and Q4. Upon submission of Q4 the page crashes: "list index out of range" (related to the pagination)
> * (minor issue) Regardless of your choices, the question counter never seems to be beyond "question 2 out of 4"

unable to recreate this, and I'm able to submit the survey successfully.

geoo89:
> unable to recreate this, and I'm able to submit the survey successfully.

It appears there's another requirement: Q2 needs to have a "go to end of survey" option.

PR #516, opened 2021-08-31
adds free text poll migration

### LINKED ISSUE #496: v1 poll free text questions migration to v2
In v1 there are poll free text questions that are without any choices
but in v2 we don't support questions like those

should we support single/multi-line text in poll 
or 
should we ignore these questions (keeping in mind if we choose to ignore then v1 submissions will also be ignored for these polls)

sheralim012:
fixes #496 

- adds free text poll content migration
- adds multiline field type in poll
- makes choices field in poll nullabel


geoo89:
This looks good. Is there a particular reason that you choose multi-line over single line here?

sheralim012:
single-line has a limit of 255 characters but multiline doesn't have this limit.

geoo89:
I see. I was thinking that I'd expect poll responses to be short, and often multiple people responding the same answer.
So there's no character limit in v1? If there is no limit, I agree with you that multi-line is safer and the correct to go.

PR #535, opened 2021-09-03
adds static content translations

### LINKED ISSUE #519: Support for customizable site strings
As we are currently lacking translations for a variety of strings, we should provide a CMS page that allows site admins to provide translations for these strings in each locale that the site offers (except English?).

Mechanism:
- If provided, use the given translated string across the site
- If not provided, use the django translated string (which defaults to English if we don't have a translation for it in django)
- If possible, pre-populate those fields where we have a translation in the given locale (e.g. django already comes with a variety of available strings).

We also need to devise a mechanism for tagging such strings in the code, or specifying the exact list of these strings.

sheralim012:
fixes #519 

- adds Translation menu
- first load translations by clicking "Load Translations" button
- now you can see all available translations and can edit those
- adds missing translation filter
- adds limited translatable string filter (on by default)

for reference:
* python manage.py makemessages -l lang-code -d django,djangojs
* python manage.py compilemessages


geoo89:
This is looking pretty good already!

Three issues:
- This lists all strings for each language, including those from django/wagtail and those used only in the backend. Can we filter the "original" column to only include strings from a fixed set of strings?
- When I update a string, it updates our corresponding PO file, but this doesn't reflect in the actual site until `manage.py compilemessages` is run. Should we have a button that invokes this, or can we invoke this automatically somehow?
- Karakalpak doesn't appear in the list of languages to filter by, probably because its language code has three letters (kaa). [You can test this by adding an entry to the .po file for kaa; while it makes other languages that are not in django/wagtail appear in the menu, this doesn't work for kaa.

sheralim012:
> This is looking pretty good already!
> 
> Three issues:
> 
> * This lists all strings for each language, including those from django/wagtail and those used only in the backend. Can we filter the "original" column to only include strings from a fixed set of strings?

yes, we can do that but that would complicate adding new strings and changing existing strings because we will be changing in code and doing a deployment after

> * When I update a string, it updates our corresponding PO file, but this doesn't reflect in the actual site until `manage.py compilemessages` is run. Should we have a button that invokes this, or can we invoke this automatically somehow?

I have already taken care of compilation on entity edit/update, could you verify again

> * Karakalpak doesn't appear in the list of languages to filter by, probably because its language code has three letters (kaa). [You can test this by adding an entry to the .po file for kaa; while it makes other languages that are not in django/wagtail appear in the menu, this doesn't work for kaa.

the package was changing lang with more than 2 characters, handled that case

geoo89:
> yes, we can do that but that would complicate adding new strings and changing existing strings because we will be changing in code and doing a deployment after

Please add a filter that filters by a fixed set of strings that are provided in the code. This filter should be enabled by default.
Ideally add another filter to only show entries that are missing their translation.

sheralim012:
> > yes, we can do that but that would complicate adding new strings and changing existing strings because we will be changing in code and doing a deployment after
> 
> Please add a filter that filters by a fixed set of strings that are provided in the code. This filter should be enabled by default.

done

> Ideally add another filter to only show entries that are missing their translation.

done

ehmadzubair:
👍🏼

geoo89:
Thanks, I think this works as intended now! I'll update the po files and list of strings in the near future.

Two minor things:
- could you put the list of strings into a separate files (so I can easily autogenerate it)?
- The filter "By limited translatable strings" is enabled by default, but in the filter menu "ALL" is selected when you initially load the page, which is a bit confusing. Could you fix that?

geoo89:
I just realized there is another issue: As this tool is updating the po files, how does this interfere with deployment updates?
For example, assume the country offices made some changes, and now we're pushing an update that adds a new string to the site (and the corresponding PO files). How would a merge of the changes be achieved?

Instead of updating the PO files for our main project, would it be possible to have this tool write to its own set of PO files, which are initially empty, but which take priority over any other PO files?

sheralim012:
> Thanks, I think this works as intended now! I'll update the po files and list of strings in the near future.
> 
> Two minor things:
> 
> * could you put the list of strings into a separate files (so I can easily autogenerate it)?

done

> * The filter "By limited translatable strings" is enabled by default, but in the filter menu "ALL" is selected when you initially load the page, which is a bit confusing. Could you fix that?

done

sheralim012:
@geoo89 

* added a management command to load po files and update db then publish all changes to po files
* this management command will automatically run on `make up`

note: we have patched loading of po file to db, update translation string from po file to db (by default the translation strings aren't updated)

home/views.py
```diff
@@ -73,11 +74,8 @@ def get_success_message(self, instance):
 
 class LoadTranslationsFromPOFiles(View):
     def get(self, request):
-        from .wagtail_hooks import TranslationEntryAdmin
-
-        manager = Manager()
-        manager.load_data_from_po()
+        call_command('load_po_files')
```
geoo89, line 77: Should this be `reload_po_files`? The way it is it crashes for me right now.

geoo89 (commented):
Running `python manage.py reload_po_files` seems to erase all changes that were made in the CMS.

Steps to reproduce:
1. Load translations in CMS (currently only xy locale has entries)
2. Change an entry
3. Reset `locale/xy/LC_MESSAGES/django.po` (this simulates an update to the codebase)
4. run `python manage.py reload_po_files` 

The previously changed entry has been reset now.

Another issue is that because the .po files are tracked by git but also changed during runtime, git will complain whenever we want to pull an update. This might not be an issue depending on how the deployment images are built.

geoo89:
Both in the PO files that we ship ("our translations") and the translation provided by the country offices ("DB"), an entry can be "complete" (has phrase + translation), "blank" (has phrase but only empty translation) or "missing" (phrase is not present).

Desired behavior of how we merge entries when we push a new update (i.e. update our translations):
- If our translation has a complete entry, update the DB with that. (This may overwrite a translation from the DB)
- If our translation has a blank entry:
    - if the DB has a translation for this entry: Use the translation from the DB
    - otherwise: store a blank entry
- If the DB has a blank entry which is missing from our translations: remove it
- If the DB has a complete entry which is missing from our translations: 
    - If it's possible to mark entries as obsolete, do that (I saw PO files containing entries being commented out). Otherwise: remove it

geoo89:
- Also write DB entries with empty translation back to PO file
- During `reload_po_files`, remove DB entries whose translation is empty if they are also missing from PO file

sheralim012:
> * Also write DB entries with empty translation back to PO file
> * During `reload_po_files`, remove DB entries whose translation is empty if they are also missing from PO file (check is for isnull rather than "" currently: https://github.com/unicef/iogt/pull/535/files#diff-66173ae9b047310006a18ed7d5f7d1ea7921273d13b2943fa48a03b32bf57c34R108)

done

geoo89:
Translations that have been removed from the PO file, but are in the DB, all get removed during `reload_po_files`.
Only those with an empty translation string should get deleted; the others should be kept.

sheralim012:
> Translations that have been removed from the PO file, but are in the DB, all get removed during `reload_po_files`. Only those with an empty translation string should get deleted; the others should be kept.

only removing entry if empty in db and not given in PO file

PR #536, opened 2021-09-03
map renamed locales in content migration

### LINKED ISSUE #404: Renamed locales for migration from v1 to v2
PR #396 renames two locales from a non-standard code used in V1 to the ISO code.
These are (see [commit](https://github.com/unicef/iogt/pull/396/commits/2a55470fad3dfaab4adc364cbbcefbe87dc56ca5)):
- Shona: sho -> sn
- Chichewa: ch -> ny

Once this PR is merged, the content migration script should be mindful and remap these as well.

### LINKED ISSUE #519: Support for customizable site strings

sheralim012:
fixes #404 

- map renamed locales to iso locales in content migrations

### DIFF
--- a/iogt_content_migration/management/commands/load_v1_db.py
+++ b/iogt_content_migration/management/commands/load_v1_db.py
@@ -247,6 +247,7 @@ class Command(BaseCommand):
             for row in section_page_translations:
                 section = self.v1_to_v2_page_map.get(self.page_translation_map[row['page_ptr_id']])
                 locale, __ = Locale.objects.get_or_create(language_code=row['locale'])
+                locale = self._get_iso_locale(locale)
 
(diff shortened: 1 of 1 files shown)


ehmadzubair:
👍🏼

PR #538, opened 2021-09-03
Fix locale mapping

geoo89:
Fix: Migrating DBs with ch/sho locales would create these in our DB, but there they are invalid (renamed to ny/sn) and crash the page
Fix: Migration would crash for surveys with images.


sheralim012:
LGTM 👍🏽

PR #551, opened 2021-09-13
fixes language switcher for non wagtail pages

### LINKED ISSUE #457: Language select dropdown only works for wagtail pages (e.g. articles)
The language select dropdown changes the language of an article/section (i.e. wagtail content page) to the selected language.

However, for non-content pages (search, profile, chatbot, etc) it simply reloads the page in the same language, regardless of which language has been selected.

Phrasing from duplicate #501:
> Picking another language that the current language on non-wagtail pages just redirects to the original url. The user should be redirected to the correct language on selecting it from the language picker

sheralim012:
fixes #457 

- fixes bug for language switcher for non-wagtail pages

### DIFF
--- a/home/templates/home/tags/language_switcher.html
+++ b/home/templates/home/tags/language_switcher.html
@@ -17,13 +17,11 @@
         {% else %}
             {% for locale in default_locales %}
                 {% get_language_info for locale.language_code as lang %}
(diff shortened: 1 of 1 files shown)


ehmadzubair:
👍🏼

PR #552, opened 2021-09-13
refactor login button

### LINKED ISSUE #485: Log in button when already logged in
Regardless of whether you're already logged in, there is a button that reads "Log In / Create Account" (or "Log In" on mobile).

- [x] When logged in, display the text "Profile" instead
- [x] When not logged in, have consistent text between mobile and desktop view (i.e. display "Log In / Create Account" in both)

sheralim012:
fixes #485 

- use consistent text for the login button
- for logged-in users show profile button


ehmadzubair:
👍🏼

PR #557, opened 2021-09-13
Fixes Comment Deletion by Moderators

### LINKED ISSUE #442: As a Moderator, I should be able to delete a comment from the IoGT frontend

### LINKED ISSUE #531: Don't show removed comments (temporarily)
Rather than displaying "This comment has been removed.", do not show the comments that have been deleted by moderators at all.
This is a temp fix as we don't have translations for this, so maybe keep comments lines to make this easy to undo.

ehmadzubair:
Moderators should now be able to delete comments using the button below:

Screenshot 2021-09-13 at 6 03 56 PM

Closes #531
Closes #442


geoo89:
Which roles are allowed to remove comments?

When testing, admins could remove comments, but they could not flag (I assume that's ok).
Editors/Moderators cannot remove comments, but can report. (Mods should be allowed to remove).
No-one can see how many users have reported a given comment (at least admins used to be able to see that?)

ehmadzubair:
Can we get on a call to discuss this?

geoo89:
I created a user which I added to the "Moderators" group, and in that group I checked "Can moderate comments" and "Object permissions - comments - delete". The latter enabled that user to delete comments.
However, even though I checked "Can moderate comments", the moderator tag doesn't appear next to that user's posts, and the "X users reported this comment" doesn't show either. I see the logic in `comments/templates/django_comments_xtd/comment_tree.html`  that should enable this, do you know why this isn't working?

PR #560, opened 2021-09-14
Add comment pagination

### LINKED ISSUE #527: Paginated Comments Listing
django-comments-xtd doesn't support pagination yet

Provide a "More" button that displays more comments (akin to chatbot).

ehmadzubair:
Closes #527


geoo89:
- Replace "Load more comments" with "More" (due to available translations)

Other than that, I only noticed a few minor issues. Fix them if they are easy, otherwise let's create issues and defer them to later:
- When you press "Load more comments", add an anchor to the newest revealed comment (so you don't have to scroll down the entire page)?
- When you reply to a comment that's only gets revealed after pressing "Load more comments", you won't see your comment (nor the comment you replied to) after replying until you press "Load more comments" again
- You display 5 root level comments, regardless of how many children they have, so even the first page could contain a lot of comments.

ehmadzubair:
> When you press "Load more comments", add an anchor to the newest revealed comment (so you don't have to scroll down the entire page)?

I think it's better if we add a new issue for the first issue.

> When you reply to a comment that's only gets revealed after pressing "Load more comments", you won't see your comment (nor the comment you replied to) after replying until you press "Load more comments" again

Fixing the second issue

> You display 5 root level comments, regardless of how many children they have, so even the first page could contain a lot of comments.
What do you suggest for the third point?

PR #561, opened 2021-09-14
fixed dropdown visibility issues

### LINKED ISSUE #459: Language picker hidden behind section image
In mobile view, when viewing a section with a lead image, the languages in the language picker are hidden behind that image.
Note that this doesn't not happen for lead images of articles, only for sections.

image

The lower one is also hidden behind the search bar in mobile:

image

sheralim012:
fixes #459 

- fixes language picker hide behind section image

### DIFF
--- a/home/static/css/global/global.css
+++ b/home/static/css/global/global.css
@@ -515,7 +515,7 @@
   font-size: 12px;
   line-height: 16px;
   color: #303030;
-  z-index: 9;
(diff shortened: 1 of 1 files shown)


sheralim012:
note: clear cache

geoo89:
The lower language picker is still hidden behind the search bar (see second part of #459).

PR #579, opened 2021-09-20
show warning of facebook free basics

### LINKED ISSUE #542: Warning when adding content incompatible with Facebook Free Basics
As a PLATFORM ADMINISTRATOR I want to be reminded when content I'm adding to the site is outside of Facebook Free Basics specifications.

This can be
1. Video
2. Large images (> 200KB)
3. Documents (which will be linked)

Should be a small notification, eg helptext, in red.

Full guidelines are here:
https://developers.facebook.com/docs/internet-org/platform-technical-guidelines
https://developers.facebook.com/docs/internet-org/how-to-submit

sheralim012:
fixes #542 

- show warning on videos add/edit
- show warning on images larger than 200kb add/edit
- show warning on documents larger than 200kb add/edit

### DIFF
--- a/iogt/static/js/iogt.js
+++ b/iogt/static/js/iogt.js
@@ -13,6 +13,18 @@ function validateFileUpload(fileInput, file_size_threshold) {
 	return true;
 }
 
+function validateFreeBasicsFileUpload(fileInput, file_size_threshold) {
+	if (!fileInput.files || !fileInput.files[0])
+		return true;
+	else {
+		var file = fileInput.files[0];
+		if (file.size >= file_size_threshold)
+			alert(`File size exceeds facebook free basics limit (${file_size_threshold / 1024}KB).`);
+	}
+
+	return true;
+}
+
 /** DOM Variables **/
 const $selects = document.querySelectorAll('.js-select');
 const $searchBtn = document.querySelector('.js-search-btn');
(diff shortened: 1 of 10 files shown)


geoo89:
image

Similarly you can add a media block to an article, and upload audio/video there.

These cases should be covered as well.

sheralim012:
> I think there are some ways of uploading media which are not covered yet:
> When editing a page, and e.g. adding a lead image or an image block for an article, you can either select from the image bank or upload a new one. If you upload a new one in this dialog, no warning is shown

try to clear cache, as this is already handled

> 
> image
> 
> Similarly you can add a media block to an article, and upload audio/video there.

handled this case too.

> 
> These cases should be covered as well.

geoo89:
Thanks, the messages are there now everywhere.

In #542 Chris clarified that documents are not allowed at all for facebook free basics, so please adjust accordingly.

Also minor issue, but sometimes the limit is displayed as 204800KB rather than 200 KB in the error message?

sheralim012:
> Thanks, the messages are there now everywhere.
> 
> In #542 Chris clarified that documents are not allowed at all for facebook free basics, so please adjust accordingly.

done

> Also minor issue, but sometimes the limit is displayed as 204800KB rather than 200 KB in the error message?

done

PR #473, opened 2021-08-17
Feat/users migration

ehmadzubair:
Fixes #397 

This PR completes the user specific migration for the IoGT data.

## Instructions for testing
Run the `load_v1_data` prior to running the `load_v1_users` command

PR #582, opened 2021-09-22
Users migration is now complete

ehmadzubair:


geoo89:
Could you address the comments from this review please (I believe only the submit date and id clash were fixed?)?

https://github.com/unicef/iogt/pull/473#pullrequestreview-749216072

ehmadzubair:
@geoo89 I've added the feedback changes and tested this with rke latest backup

PR #622, opened 2021-10-07
User migrations fixes (Nested comments + Duplicate users)

### LINKED ISSUE #588: Handling duplicate user names during migration
Before running user migrations, one (or possibly multiple) users are created for the site that can be used by the country offices to populate the site with content. It is possible that the user names of one of those users coincides with a username from a migrated account (e.g. a common case could be 'admin'). How should we deal with this?

- Keep the v1 user
- Keep the v2 user
- Keep both and rename the v1 user
- Keep both and rename the v2 user

### LINKED ISSUE #593: Migrating nested comments
> Can we keep existing nested comments (hidden behind view more) however don’t allow new comments to be nested past depth of 1. Change view if users want to see all nested comments

ehmadzubair:
Didn't encounter a situation where the comments were nested on multiple levels (am i missing something?). This fixes the migration to preserve the structure. 

Closes #593 
Closes #588

### DIFF
--- a/iogt_content_migration/management/commands/load_v1_users.py
+++ b/iogt_content_migration/management/commands/load_v1_users.py
@@ -190,21 +190,33 @@ class Command(BaseCommand):
             user_data = dict(row)
             user_data.update({'has_filled_registration_survey': True})
 
-            user, __ = get_user_model().objects.update_or_create(username=row['username'], defaults=user_data)
-            V1ToV2ObjectMap.create_map(user, v1_user_id)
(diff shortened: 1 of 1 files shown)


geoo89 (approved):
It appears that comments in v1 cannot be nested beyond depth 1 either: Neither as a user on a page, nor as admin in the comments panel can I reply to comments that are at depth 1. The migration seems to work well for comments now.

I realized that in v1 user names are case sensitive, while in v2 they are not, which causes some issues (e.g. in RKE there are admin and ADMIN users). I made a new issue for this: #627

Issue #640, opened 2021-10-11
Migrate page button blocks


geoo89:
V1 articles (also survey descriptions) can have blocks of type "page". These are currently not migrated. (Is the block type name different in V2?)

PR #647, opened 2021-10-12
update questionnaires description field and fixes content migration

### LINKED ISSUE #639: Survey description content blocks
- v1 survey descriptions support heading/paragraph/image/list/numbered list/page
- v2 only supports paragraph and image

TODO:
- Extend v2 to support these additional block types
- Fix migration code to ensure these blocks are ported over.

sheralim012:
fixes #639 

- add fields (heading, paragraph, image, list, numbered list, page) to the questionnaires description field
- fixes content migration for the survey description field


questionnaires/models.py
```diff
@@ -57,8 +57,12 @@ class QuestionnairePage(Page, PageUtilsMixin):
 
     description = StreamField(
         [
-            ("paragraph", blocks.RichTextBlock()),
+            ('heading', blocks.CharBlock(form_classname="full title")),
+            ('paragraph', blocks.RichTextBlock(features=settings.WAGTAIL_RICH_TEXT_FIELD_FEATURES)),
             ("image", ImageChooserBlock()),
+            ('list', blocks.ListBlock(blocks.CharBlock(label="Item"))),
+            ('numbered_list', blocks.ListBlock(blocks.CharBlock(label="Item"))),
+            ('page', blocks.PageChooserBlock()),
```
geoo89, line 65: Probably better to be consistent with the blocks in `home/models.py` and use `page_button`?

iogt_content_migration/management/commands/load_v1_db.py
```diff
@@ -977,22 +977,22 @@ def create_survey(self, v2_index_page, row):
...
         for block in v1_survey_description:
-            if block['type'] == 'paragraph':
+            if block['type'] in ['heading', 'paragraph', 'list', 'numbered_list']:
                 v2_survey_description.append(block)
             elif block['type'] == 'image':
                 image = self.image_map.get(block['value'])
                 if image:
                     v2_survey_description.append({'type': 'image', 'value': image.id})
+            elif block['type'] == 'page':
+                page = self.v1_to_v2_page_map.get(block['value'])
```
geoo89, line 993: I don't think this works, as `v1_to_v2_page_map` is populated as we go through the pages. So if a survey links to a page that hasn't been processed yet, the lookup will fail even though it shouldn't.
Also see #640 (maybe you can fix this as part of this PR as well).

sheralim012:
@geoo89 

updated migrations so before pull run `python manage.py migrate questionnaires 0019` and then pull

PR #664, opened 2021-10-15
adds page link

### LINKED ISSUE #663: Add a PageLink Page to Footer Index Page
Supported Pages:
  - Add Icon field to all questionnaire pages

Fields:
- Title
- Page (FK) - limited to [Section, Article, Survey, Poll, Quiz]

This page can only be the child of FooterIndexPage. Display these pages like FooterPage buttons on the frontend. The icon of the linked page will be used as the icon in the footer button. Update the implementation of the Footer section to fetch the buttons from the Footer index page so that they're sorted exactly like they are in the admin.

sheralim012:
fixes #663 

- adds page link page
- adds icons in questionnaires


home/models.py
```diff
@@ -459,9 +463,29 @@ class FooterPage(Article):
...
+class PageLinkPage(Page):
+    parent_page_types = ['home.FooterIndexPage']
+    subpage_types = []
+
+    page = models.ForeignKey(Page, related_name='page_link_pages', on_delete=models.PROTECT)
+
+    content_panels = Page.content_panels + [
+        PageChooserPanel('page')
+    ]
+
```
ehmadzubair, line 482: Add a one line comment for ducktyping

home/models.py
```diff
@@ -448,7 +448,11 @@ class BannerPage(Page):
 
 class FooterIndexPage(Page):
     parent_page_types = ['home.HomePage']
-    subpage_types = ['home.FooterPage']
+    subpage_types = ['home.FooterPage', 'home.PageLinkPage']
```
geoo89, line 451: Could you add support for the other page types (article, survey, poll, quiz) as well?

geoo89 (commented):
Is it possible to inherit the title from the page that is referenced, rather than adding it manually?
If this is very tricky, it's not strictly necessary though.

(Later maybe an optional title/icon field could be used to overwrite the title/icon of the referenced page).

home/models.py
sheralim012: done

PR #669, opened 2021-10-18
:zap: Added commenting options

### LINKED ISSUE #584: Article comment status
Currently: open or closed

in V1: Open, closed, disabled, timestamped.

closed vs disabled: closed shows past comments, disabled does not
timestamped: time window when commenting is possible is provided

ehmadzubair:
Closes #584 

Now you should be able to both set the commenting statuses and see them migrated from v1 data.

Note: The datetime field is in GMT timezone as with all the wagtail admin datetime fields. Should we just let them be or find a workaround to that?

### DIFF
--- a/comments/models.py
+++ b/comments/models.py
@@ -42,11 +42,14 @@ class CommentableMixin(models.Model):
         return self.commenting_status in [CommentStatus.OPEN, CommentStatus.CLOSED, CommentStatus.TIMESTAMPED]
 
     def should_show_new_comment_box(self):
-        commenting_still_valid = self.commenting_starts_at > timezone.now() > self.commenting_ends_at
+        commenting_still_valid = True
+        if self.commenting_starts_at:
+            commenting_still_valid = self.commenting_starts_at > timezone.now()
+        if self.commenting_ends_at:
+            commenting_still_valid = commenting_still_valid and self.commenting_ends_at < timezone.now()
 
         return self.commenting_status == CommentStatus.OPEN or \
-               (self.commenting_status == CommentStatus.TIMESTAMPED and self.commenting_starts_at
-                and self.commenting_ends_at and commenting_still_valid)
+               (self.commenting_status == CommentStatus.TIMESTAMPED and commenting_still_valid)
 
     def get_absolute_url(self):
         if isinstance(self, Page):


iogt_content_migration/management/commands/load_v1_db.py
```diff
@@ -678,6 +692,7 @@ def create_footer(self, row):
             last_published_at=row['last_published_at'],
             search_description=row['search_description'],
             seo_title=row['seo_title'],
+            commenting_status=CommentStatus.CLOSED
```
geoo89, line 695: V1 footers also have `commenting_starts_at` and `commenting_ends_at`.

iogt_content_migration/management/commands/load_v1_db.py
```diff
@@ -435,6 +436,7 @@ def migrate_articles(self):
             if row['page_ptr_id'] in self.page_translation_map:
                 article_page_translations.append(row)
             else:
+
                 self.create_article(row)
         else:
             for row in article_page_translations:
```
geoo89, line 442: Shouldn't these fields also be populated for translated articles?

iogt_content_migration/management/commands/load_v1_db.py
ehmadzubair: Yes, but they are nullable, so we'll just leave them as is. right?

geoo89: But this data is actually populated in V1, so we should port it over.

ehmadzubair:
@geoo89 addressed comments. Good catch. Thank you

comments/models.py
```diff
@@ -1,23 +1,53 @@
...
+        FieldPanel('commenting_starts_at', heading='Commenting Opens At'),
+        FieldPanel('commenting_ends_at', heading='Commenting Ends At')
+
     ]
 
+    def should_show_comments_list(self):
+        return self.commenting_status in [CommentStatus.OPEN, CommentStatus.CLOSED, CommentStatus.TIMESTAMPED]
+
+    def should_show_new_comment_box(self):
+        commenting_still_valid = self.commenting_starts_at > timezone.now() > self.commenting_ends_at
```
geoo89, line 45: This throws an error if one of the time fields is left blank.
I'd suggest something like this (still needs fix to do timezone-aware/naive comparison):

```
        commenting_starts_at = self.commenting_starts_at or datetime.datetime.min
        commenting_ends_at = self.commenting_ends_at or datetime.datetime.max
        commenting_still_valid = commenting_starts_at < timezone.now() < commenting_ends_at
```

This way, e.g. leaving the commenting_ends field blank means commenting stays open indefinitely.

PR #742, opened 2021-10-28
updates featured article in homepage

### LINKED ISSUE #685: Add option to show Section titles etc on Article content blocks
This is most relevant on the homepage (not sure if Article content blocks appear elsewhere).

V1 included Section titles on the homepage:

image

It would be useful to have the option to show users the section the Article appears in, as Page titles are often written "within the context" of a certain section.

sheralim012:
fixes #685 

- updates article block for home page featured articles
- updates content migration to use new featured articles block
- updates UI for home page featured articles
- set display_section_title=True in content migration
- added data migration to fix existing homepage featured content, and articles


geoo89:
Thanks, it seems like the migration is working well and old pages get migrated properly.

Only small issue is that the section text that's displayed above the articles should be from the top-level section (e.g. Covid-19, Youth) rather than the parent section, and if easy to do be displayed in the color of that section.

I'm happy to merge this as it is though and leave that for a separate issue. So @ehmadzubair if you don't find any issues while reviewing this feel free to merge it.

PR #765, opened 2021-11-02
Prevent new user registrations that differ only in capitalization

### LINKED ISSUE #654: Prevent new user registration that differ only in capitalization.
E.g. If a user 'admin' exists, it should be impossible to register as 'Admin' or 'ADMIN'.

Rationale: #627 and #631

falsefruit:
Closes #654 

Prior to making changes for this I noted that from the iogt app, a user can NOT register a new user with a username that differs only in terms of case. For example, I created a user foobar:

image
---
image
---
I then attempted to create a user FoObAr, and was refused with the following validation message:

image
---
But, a user CAN do it from the Wagtail admin interface. This change adds validation to `WagtailAdminUserCreateForm` to mirror the above behavior:

image


geoo89:
Thank you for comprehensively investigating this (good to think of the admin panel as well), and providing a fix!

PR #766, opened 2021-11-02
Fix scaling factor when calling cairosvg.svg2png

### LINKED ISSUE #719: Some svg-to-png icons are too large
Example: http://mena.iogt.site/media/svg-to-png-maps/svg-to-png_PpUYBHA.png

image

Unnecessarily large PNG for an icon causing large file size.

Marking urgent because I think these PNGs are populated only once.

falsefruit:
Fixes #719 

In `SVGToPNGMap.get_png_image()`, a scaling factor of 10 is being passed to `convert_svg_to_png_bytes()`, resulting in a 10x scaling of the resulting png image when that scale value is passed to `cairosvg.svg2png()`. Additionally, the default scale value in `convert_svg_to_png_bytes()` is 100, which would result in a 100x scaling. Setting this to 1 results in a png that is the same width and height as the source svg.

As an example, I downloaded a [freely available svg file](https://www.svgrepo.com/svg/365021/aperture-thin) which has a defined width and height of 256px. I uploaded it into the iogt app before and after making this change. The below screenshot shows that before making this change, the resulting png is scaled 10x, or 2560x2560 px (195 kB):

image
---
Whereas after this change, the resulting png is the original size of 256x256 px (14.3 kB):

image
---
Note: I believe after merging this, any existing SVGs will need to be re-uploaded to be resized properly.

### DIFF
--- a/home/utils/image.py
+++ b/home/utils/image.py
@@ -6,7 +6,7 @@ from bs4 import BeautifulSoup
 from django.core.files.base import ContentFile
 
 
-def convert_svg_to_png_bytes(svg_file_path, fill_color=None, stroke_color=None, scale=100):
+def convert_svg_to_png_bytes(svg_file_path, fill_color=None, stroke_color=None, width=32):
     with open(svg_file_path, 'r', encoding='utf-8') as f:
         soup = BeautifulSoup(f, 'lxml')
         elements_to_fill = soup.find_all('path')
@@ -19,5 +19,5 @@ def convert_svg_to_png_bytes(svg_file_path, fill_color=None, stroke_color=None, 
     parsed_soup = str(soup.body.next)
     parsed_soup = parsed_soup.replace('viewbox', 'viewBox')
 
-    file_bytes = cairosvg.svg2png(bytestring=parsed_soup, scale=scale)
+    file_bytes = cairosvg.svg2png(bytestring=parsed_soup, output_width=width)
     return ContentFile(file_bytes, 'svg-to-png.png')
--- a/home/models.py
+++ b/home/models.py
@@ -984,7 +984,7 @@ class SVGToPNGMap(models.Model):
         try:
             obj = cls.objects.get(svg_path=svg_path, fill_color=fill_color, stroke_color=stroke_color)
         except cls.DoesNotExist:
-            png_image = convert_svg_to_png_bytes(svg_path, fill_color=fill_color, stroke_color=stroke_color, scale=10)
+            png_image = convert_svg_to_png_bytes(svg_path, fill_color=fill_color, stroke_color=stroke_color, width=32)
             obj = cls.objects.create(
                 svg_path=svg_path, fill_color=fill_color, stroke_color=stroke_color, png_image_file=png_image)
         return obj.png_image_file


cbunicef:
Thank you @falsefruit! I think in this case we would rather have the PNG sized to the desired dimensions, rather than the same as the original .svg, as our SVGs normally (always?) are icons. The smaller size will save bandwidth and also look better resized from SVG than resized in browser while PNG.

@geoo89 do you agree? Do we have a single icon size that could be the output?

geoo89:
I agree, currently our only use case is the icons. Their display size seems to be (as indicated by 'rendered size' above) 16x16, @azahid89 or @ehmadzubair can you confirm?
Thus I think 16x16 would be a reasonable target size.

> Note: I believe after merging this, any existing SVGs will need to be re-uploaded to be resized properly.

It should be sufficient to clear the cache of the generated PNGs (which can be done in django-admin), so that they are recreated in the smaller size.

cbunicef:
My reading of the manual for cairosvg is that we can specify height and width output (or probably both). Let's choose one or the other so that if an icon is not perfectly square it doesn't stretch. Which would be better? My guess is better to specify height.

ehmadzubair:
16px by 16px is the rendered size of the icon. I think the 256 x 256 intrinsic size should be good. 16px intrinsic size will result in a very pixelated logo

geoo89:
Does the conversion not use anti-aliasing? If it does, why would it look much different displaying a 16x16 px image compared to a 256x256 image that is displayed as 16x16 size?

> Which would be better? My guess is better to specify height.

My intuitive guess here would have been width: vertically, the footer/section buttons can differ in size e.g. if the text doesn't fit on a single line, and so the alignment should still look fine if an icon and thus button is taller vertically, but horizontally a wider icon would misalign the text compared to other buttons?

cbunicef:
> My intuitive guess here would have been width: vertically, the footer/section buttons can differ in size e.g. if the text doesn't fit on a single line, and so the alignment should still look fine if an icon and thus button is taller vertically, but horizontally a wider icon would misalign the text compared to other buttons?

Yes agreed, let's fix the width.

falsefruit:
I have updated the PR to utilize a fixed width of 16px.

Below is a screenshot of the same svg-to-png image rendered at 16x16 by Chrome 95 on Linux under the following circumstances:

1. Existing behavior (10x scaling)
2. My original PR (1x scaling)
3. Changed to 16px intrinsic width

image

---

So at default 100% scaling, it looks perfectly fine. If the user is utilizing any zoom, it will start to look bad. For example, this is with Chrome set to 150% zoom:

image

Similar visual quality issues may occur on HiDPI devices, but I'm not sure.

---

From a bandwidth perspective, the 16px intrinsic width does result in a significant additional reduction in size in this example:

Existing behavior: 195 kB
My original PR: 14.3 kB
16px intrinsic width: 588 B

So if bandwidth is at an absolute premium, the possible issues with this approach may be acceptable.

cbunicef:
Thanks @falsefruit, this analysis is greatly appreciated.

Every KB does count, but you are right that we need to balance it. It's likely some users will be zooming on Chrome or other browsers as an accessibility tool.

Am I correct in understanding that if we scale to 32x32 we will accommodate 200% zoom without stretching? I imagine we would net out around 2KB which is very acceptable.

geoo89:
Thank you for testing this out and providing the detailed analysis, this is extremely helpful!

>  If the user is utilizing any zoom, it will start to look bad. [...] Similar visual quality issues may occur on HiDPI devices, but I'm not sure.

@cbunicef What is your take on this? Basically, we want to avoid displaying images at a size larger than their intrinsic size.
I don't know how commonly this will occur for 16x16 in practice... We could also go for some intermediate size, like 32x32 or 64x64 to get a smaller bandwidth than the ~14.3 kB we get for 256x256 while still looking decent for moderate zoom levels.

home/utils/image.py
```diff
@@ -6,7 +6,7 @@
 from django.core.files.base import ContentFile
 
 
-def convert_svg_to_png_bytes(svg_file_path, fill_color=None, stroke_color=None, scale=100):
+def convert_svg_to_png_bytes(svg_file_path, fill_color=None, stroke_color=None, width=16):
```
sheralim012, line 9: update the usage of this parameter in the function

sheralim012 (changes requested):

home/utils/image.py
cbunicef: This should be 32 rather than 16 right?

geoo89: Yes. I believe the line highlighted in green is the current state of the code, not the suggested change.

falsefruit: Sorry for the delay, I've updated this.

falsefruit:
I've updated this such that the output png is 32px wide, and confirmed that it renders nicely up to 200% zoom. In the below screenshot, #3 has the 16px wide image, and #4 has the 32px wide one:

image

PR #785, opened 2021-11-09
add sorting to load v1 db

### LINKED ISSUE #709: During migration, different IoGT V1 sites should migrate Pages order in Sections differently
Due to various versions of IoGT V1 currently in use at UNICEF, and outstanding bugs, there are two different ways that Pages were ordered in Sections:

Type 1:
kenya
tanzania
south africa
eucador
IOGT global

Pages in Sections were ordered by Creation Date. That order of Pages should be transferred to V2 during migration. eg

V1 newest creation date = V2 first item
V1 second newest creation date = V2 second item

Type 2:
All others

Pages in Section were ordered via user-selected order - which is NOT the same as order which appears by default in the Admin Panel interface. It is the order which appears after clicking the "Sort" button at the Section level. That order of Pages should be transferred to V2 during migration. eg

V1 first item when sorting Pages = V2 first item
V1 second item when sorting Pages = V2 second item

sheralim012:
fixes #709 

- removed skip-locales flag from load_v1_db
- added sort flag which expect `type1` or `type2`
- if type1 is given then sort by first_published_at
- if type2 is given then sort by path
- articles are sorted within sections
- page link pages are added after articles within section and added report for it


iogt_content_migration/management/commands/load_v1_db.py
```diff
@@ -1649,6 +1637,33 @@ def move_page(self, page_to_move, position):
...
+        for page in pages:
+            children = page.get_children().specific()
+            children_list = []
+            for child in children:
+                v1_id = V1ToV2ObjectMap.get_v1_id(child, child.id)
+                if v1_id:
+                    cur = self.db_query(f'select * from wagtailcore_page wcp where id = {v1_id}')
+                    v1_row = cur.fetchone()
+                    cur.close()
+                    setattr(child, 'creation_date', v1_row['latest_revision_created_at'])
```
geoo89, line 1655: use `first_published_at` instead

sheralim012: done

geoo89:
It appears that the sorting doesn't always work.

Example:
http://localhost:8000/en/sections/parents-caregivers/nutrition-and-breastfeeding/what-works-and-what-hurts/
(Compare to https://www.internetofgoodthings.org/section/parents-caregivers/nutrition-and-breastfeeding/what-works-and-what-hurts/)
I checked the DB entries and by creation date, these articles should be ordered:

- Your opinion is important to us!
- Feeding your newborn
- The role of birth attendants
- For working mothers
- During delivery

After running the migration with `--sort type1` instead I get:
- Feeding your newborn
- Your opinion is important to us!
- During delivery
- The role of birth attendants
- For working mothers

sheralim012:
@geoo89 changes addressed

geoo89:
Thank you! It appear that the specification was wrong/incomplete. See here for the full sorting that needs to be done: https://github.com/unicef/iogt/issues/709#issuecomment-969018032

geoo89:
Articles are displayed in the wrong order again. (It was working prior to 01ff8f6).
Example: https://www.internetofgoodthings.org/section/connect-smart/your-rights/know-and-defend-your-rights/

http://localhost:8000/en/sections/connect-smart/your-rights/know-and-defend-your-rights/
image

Note: the timestamps are all within a few seconds of each other, not sure if this is relevant here.

sheralim012:
@geoo89 pushed a fix for it.

PR #810, opened 2021-11-19
fixes permission inside container issue

### LINKED ISSUE #808: Permission denied error when running `reload_po_files` command
This occurs when running the app in the production docker container. 

```
iogt-django-1  | Reloading translations...
iogt-django-1  | The translations have been loaded successfully!
iogt-django-1  | Traceback (most recent call last):
iogt-django-1  |   File "manage.py", line 10, in <module>
iogt-django-1  |     execute_from_command_line(sys.argv)
iogt-django-1  |   File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
iogt-django-1  |     utility.execute()
iogt-django-1  |   File "/usr/local/lib/python3.8/site-packages/django/core/management/__init__.py", line 395, in execute
iogt-django-1  |     self.fetch_command(subcommand).run_from_argv(self.argv)
iogt-django-1  |   File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 330, in run_from_argv
iogt-django-1  |     self.execute(*args, **cmd_options)
iogt-django-1  |   File "/usr/local/lib/python3.8/site-packages/django/core/management/base.py", line 371, in execute
iogt-django-1  |     output = self.handle(*args, **options)
iogt-django-1  |   File "/app/home/management/commands/reload_po_files.py", line 20, in handle
iogt-django-1  |     manager.update_po_from_db(lang=language)
iogt-django-1  |   File "/app/iogt/patch.py", line 170, in update_po_from_db
iogt-django-1  |     pofile.save(pofile_path)
iogt-django-1  |   File "/usr/local/lib/python3.8/site-packages/polib.py", line 440, in save
iogt-django-1  |     fhandle = io.open(
iogt-django-1  | PermissionError: [Errno 13] Permission denied: '../usr/local/lib/python3.8/site-packages/django/contrib/auth/locale/ar/LC_MESSAGES/django.po'
iogt-django-1 exited with code 1
```

This can be reproduced with docker compose by using `Dockerfile` instead of `Dockerfile.dev` in docker-compose.yml.

The production container image runs as the app as the user "wagtail", which may be the cause of this problem.

sheralim012:
fixes #808 

- gives wagtail user permission for /usr/ dir
- removes compilemessages from Dockerfile as docker-compose.yml takes care of it


geoo89:
Is this safe? Or would it be more sensible to write any changes that have been done to strings from wagtail to the PO files of our project?

istride (commented):
I'd prefer it if the wagtail user is as restricted as possible. I echo Georg's question about whether there might be a different way to override translated strings that doesn't involve changing system-level files.

I think it may be best to keep running `compilemessages` at build time, just in case it is not necessary to do it at runtime.

sheralim012:
@istride  another way is to use virtual env. inside docker and give user permission for it, so system-level files wouldn't be affected by it

geoo89:
When you run reload_po_files, does it load all PO files or only those from our project now?
I cannot get the wagtail/django strings to load anymore (but for some reason I cannot get them even without this PR).
But now I'm thinking that maybe this is actually a good thing: If additionally we could only show those entries from our PO files which do not have a corresponding entry in the wagtail/django PO files (i.e. we don't show the ones which are redundant), then I could just put blank entries for those English strings where wagtail/django has translations for some locales but not all, and then the country offices would see and could populate the fields only for the missing locales.

istride (approved):
Fine from a deployment perspective. I think Georg has further questions.

PR #812, opened 2021-11-22
Add wagtail transfer configuration in production environment

istride:


iogt/settings/docker_compose_dev.py
```diff
@@ -11,7 +11,16 @@
...
         'AUTO_UPDATE': True,
         'ATOMIC_REBUILD': True
-        }
+    }
+}
+
+
+WAGTAILTRANSFER_SECRET_KEY = os.environ.get('WAGTAILTRANSFER_SECRET_KEY')
+WAGTAILTRANSFER_SOURCES = {
+   'iogt_global': {
```
geoo89, line 20: Should we also make this configurable?

geoo89 (approved):
LGTM.

@sheralim012 or @ehmadzubair please merge if you agree.

iogt/settings/docker_compose_dev.py
istride: I think we should because that is the easiest way to keep the secret key secret and changeable for each environment and site.

geoo89: I was referring to the `iogt_global` label.

istride: ah, right...

yes, I suppose we should, and I've updated the PR to reflect this.

PR #817, opened 2021-11-23
uses translated from page creation datetime for translated pages sorting

### LINKED ISSUE #811: Sort translated pages during migration
English pages within a section are sorted according to https://github.com/unicef/iogt/issues/709#issuecomment-969018032

However, this is not the case yet for translated pages: At least the section and article order is wrong. The same sorting should be applied to translated pages.

sheralim012:
fixes #811 

- translated from first_published_at date is used for translated pages

### DIFF
--- a/iogt_content_migration/management/commands/load_v1_db.py
+++ b/iogt_content_migration/management/commands/load_v1_db.py
@@ -1701,6 +1701,9 @@ class Command(BaseCommand):
                 except:
                     continue
                 if v1_id:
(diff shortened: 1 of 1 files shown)


geoo89:
Also for subsections, the order of subsections within a section should mirror the order of the corresponding subsections in English.

e.g. check http://localhost:8000/fr/sections/parents-caregivers/ The subsections are jumbled, should be in increasing order of age of the child.

sheralim012:
@geoo89 sorted translated sub sections based on translated from sub sections

PR #828, opened 2021-11-25
Enh/mark users with display name collision

### LINKED ISSUE #827: User group for display name collisions
Create a special user group during `load_v1_users` which contains those users that have a display name collision with another user __who registered before them__.
E.g. Of the set of users with the name 'John', the first registered user 'John' can keep their name and is not added to the group, but all later user 'John' will be added to the group.

The aim is to later prompt these users, when they log in for the first time, to change their display name, to ensure that no two users have the same display name.
(Instead of a group, a flag could also be used, as long as it facilitates this use case.)

sheralim012:
fixes #827 

- added a group named "Colliding Display Names"
- users with colliding display names will be added to the above-mentioned group in load_v1_users

### DIFF
--- a/comments/templates/comment_reply.html
+++ b/comments/templates/comment_reply.html
@@ -14,7 +14,7 @@
     </header>
     <div class="row nice-padding">
         <h2>Comment </h2>{{ comment.comment }}
-        <h4>User</h4>{% if comment.user_name == 'anonymous' %}Anonymous{% else %}{{ comment.user.display_name }}{% endif %}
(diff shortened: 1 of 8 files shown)


ehmadzubair (approved):

geoo89:
It appears that users whose display name is the same as their login name get added to the collision group.

This should be fixed.

geoo89:
I cherry-picked 8b14834 895d964 b7e1ccc 0eb5676 7f78bf2 e633537.

Your fix for the infinite redirection works if there are multiple site objects, but causes an infinite re-direction if there is only one (which was working previously). Thus I reverted the Site-object related commits so we just create one Site object during the migration. Will set the Site object manually for now until we figure out what to do about them.

PR #843, opened 2021-11-30
adds migration for page revisions

### LINKED ISSUE #841: During migration, Page history is lost
Do we have any way to retain this? 

It would be useful to understand, especially, when content was "Last Updated". No need to go into the Revisions tool from V1, I don't think it works.

sheralim012:
fixes #841 

- adds migration for page revision
- page revision is stored as a JSON dump and due to structure change from v1 to v2 we can't access these revisions
- these page revisions can tell when was the page last edited and by who


geoo89:
We have the "Updated" column in V2: 
image

It seems to be populated with a value called "latest_revision_created_at", which I believe V1 pages also have. Can we populate this?

sheralim012:
changes addressed @geoo89

PR #844, opened 2021-11-30
optimizes user export query

### LINKED ISSUE #842: Exporting Users times out
For example on global.iogt.site

https://global.iogt.site/admin/iogt_users/user/

When using "Download XLXS" the site pauses and then gives a 504 error.

When using "Download CSV" the site begins to download the CSV file but then stops with a "network error".

Presumably there is so much data to collect that the server times out before the data collection is complete.

It's worth also noting that the site is inaccessible for users while downloading these files, and those page loads result in 504.

Does this also happen when there is a large Form Data download?

sheralim012:
partially addresses #842 

- optimizes user export query

### DIFF
--- a/iogt_users/wagtail_hooks.py
+++ b/iogt_users/wagtail_hooks.py
@@ -1,10 +1,14 @@
+from functools import cached_property
+
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.db.models import Prefetch
 from django.utils import timezone
 from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
 
 from home.models import SiteSettings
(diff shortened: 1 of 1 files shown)


iogt_users/wagtail_hooks.py
```diff
@@ -22,17 +26,41 @@ class UsersExportAdmin(ModelAdmin):
     menu_order = 601
 
     def registration_survey_response(self, obj):
+        user_submissions = obj.user_submissions
+
+        return user_submissions[0].form_data if user_submissions else ''
+
+    @cached_property
+    def get_registration_survey_ids(self):
+        ids = []
```
ehmadzubair, line 35: check with breakpoint

PR #845, opened 2021-11-30
adds section commenting status and inheritance

sheralim012:
- adds comment related fields in the section
- by default page comment status will be inherited
- if article commenting status is inherited then ancestors will be used to determine the commenting status
- updated content migration to handle section commenting fields


ehmadzubair (approved):

comments/models.py
```diff
@@ -26,30 +29,57 @@ class CommentableMixin(models.Model):
...
+                return ancestor
+
     def should_show_comments_list(self):
-        return self.commenting_status in [CommentStatus.OPEN, CommentStatus.CLOSED, CommentStatus.TIMESTAMPED]
+        from home.models import Section
+
+        commenting_status = self.commenting_status
+        if commenting_status == CommentStatus.INHERITED:
+            ancestor = self._get_valid_ancestor_for_commenting()
+            commenting_status = ancestor.commenting_status
```
geoo89, line 58: Check for None, e.g. if the whole ancesteor tree has INHERITED, the page crashes.
In that case, use 'closed' as a default. (Or check what V1 uses, might be not so easy to find out though.)

comments/models.py
```diff
@@ -26,30 +29,57 @@ class CommentableMixin(models.Model):
...
-        return self.commenting_status == CommentStatus.OPEN or \
-               (self.commenting_status == CommentStatus.TIMESTAMPED and commenting_still_valid)
+        commenting_still_valid = True
+        commenting_status = self.commenting_status
+        commenting_starts_at = self.commenting_starts_at
+        commenting_ends_at = self.commenting_ends_at
+
+        if commenting_status == CommentStatus.INHERITED:
+            ancestor = self._get_valid_ancestor_for_commenting()
+            commenting_status = ancestor.commenting_status
```
geoo89, line 72: same here.

geoo89 (commented):
Looks good, except if the ancestor recursion bottoms out the page crashes.

sheralim012:
@geoo89 fixed

comments/models.py
sheralim012: v1 was using disabled for it

PR #876, opened 2021-12-07
Explicitly set requirements for Django and Redis

istride:
To ensure that the correct versions of Django and Redis are installed, this PR explicitly states which versions are required in the app's requirements.txt file.

The docker container could not be built because Django 4 was being installed instead of < 3.2, which is required by Wagtail.


geoo89 (approved):
LGTM

PR #878, opened 2021-12-08
creates a management command to sync translated pages

### LINKED ISSUE #868: Missing fields when translating articles after model update
This is currently happening on global. Steps to reproduce:
- Create a locale without articles
- Navigate to some article in English
- More > Translate this page > [The new locale]
- We get a warning "The data model for 'Article' has been changed since the last translation sync. If any new fields have been added recently, these may not be visible until the next translation sync."
- Body content blocks are missing from the translation
- After "Stop translation", the body content is empty

This occurs for articles that were created as part of the content migration. Note that there has been a site upgrade on global since the migration, including, among other PRs, #768 . 
For newly created articles, this does not occur.

I tested this on iogt-django-dev, with a previously created page the last two steps didn't occur, however the warning is still issued. Again, for a newly created page this issue doesn't occur.

Is this related to updates to the Article or related models, with the translation models not being updated accordingly?

Needs investigation.

sheralim012:
Fixes #868

- created a management command to sync translated pages
- added a --publish flag to automatically publish after synchronization

usage:
- `python manage.py sync_translated_pages`
- `python manage.py sync_translated_pages --publish`

### DIFF
--- a/home/management/commands/sync_translated_pages.py
+++ b/home/management/commands/sync_translated_pages.py
@@ -1,0 +1,29 @@
+from django.core.exceptions import ValidationError
+from django.core.management.base import BaseCommand
+from wagtail.core.models import Page
+from wagtail_localize.models import TranslationSource
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--publish',
+            action='store_true',
+            help='This will publish immediately, before any new translations happen.'
+        )
+
+    def handle(self, *args, **options):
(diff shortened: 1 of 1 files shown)


geoo89:
This looks good to me and appears to resolve the issue.
At this point, I think we want to run it without the --publish flag as that publishes draft pages and updates the last-edited timestamp of all pages. We have to make sure that this also works without --publish when we update the page models, or be a bit more careful about publishing vs saving as draft and not touching the time stamps.

PR #889, opened 2021-12-13
do not update po files, just return translation if found in db

### LINKED ISSUE #888: Custom translations don't take effect in production
Translations entered via translation manager don't take effect on the site in prod, only in dev.

- We are using gunicorn on prod and it doesn’t restart when PO files are updates on translation save, although there is a flag for gunicorn to do that but it’s for development purposes
- when we save translations the PO files do get updated but they don’t get synchronized on other pods in kubernetes

Solution:
- monkey patch translate/trans tag of django and use DB to return the translation if found otherwise use default behavior (will do some R&D to check the feasibility)

(Thanks Sher for the outline of the issue and solution)

sheralim012:
fixes #888

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       WAGTAILTRANSFER_SOURCE_BASE_URL: https://example.com/wagtail-transfer/
       WAGTAILTRANSFER_SOURCE_SECRET_KEY: wagtailtransfer-source-secret-key
       BASE_URL: 'http://localhost:8000'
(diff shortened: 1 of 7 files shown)


geoo89:
- If the translation string of a translation entry is "", don't use that. (Currently, a blank string is displayed then, e.g. go to the Log out page in French)
- After editing a translation, all filters are reset. Ideally keep the filters the user had enabled, or at least enable the filters that we use by default (i.e. limited translatable strings = Yes)

PR #918, opened 2021-12-16
Standalone tool to check the status of all published pages on a site

geoo89:


common/check_page_status/check_page_status.py
```diff
@@ -0,0 +1,44 @@
+from urllib.request import Request, urlopen
+import json
+import requests
+import argparse
+import logging
+
+
+def get_sitemap(url):
+    req = Request(url + 'sitemap/?format=json')
```
sheralim012, line 9: is there any specific reason you aren't using `requests.get()`

common/check_page_status/check_page_status.py
```diff
@@ -0,0 +1,44 @@
...
+    req.add_header('Accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8')
+    req.add_header('Accept-Charset', 'ISO-8859-1,utf-8;q=0.7,*;q=0.3')
+    req.add_header('Accept-Encoding', 'none')
+    req.add_header('Accept-Language', 'en-US,en;q=0.8')
+    content = json.loads(urlopen(req).read().decode('utf-8'))
+    return content
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Check status of all pages of a given IoGT site.')
```
sheralim012, line 20: if easier could you convert this to a management command?

sheralim012 (changes requested):

common/check_page_status/check_page_status.py
geoo89: Will do.

common/check_page_status/check_page_status.py
geoo89: How would I do that?

istride:
@ChrisMarsh82 I suggest closing this PR if it is no longer necessary.

PR #933, opened 2021-12-20
adds warning for page with empty image

### LINKED ISSUE #924: During migration, some articles have image blocks with no associated images
https://rw.iogt.site/admin/pages/171/edit/

image

sheralim012:
fixes #924 

- adds warning for pages with empty image


geoo89:
Blank lead images for sections or articles etc are fine. 

Where we actually need the warnings are content blocks within the article body and survey/poll description which are image content blocks but don't reference a valid image.

sheralim012:
> Blank lead images for sections or articles etc are fine.
> 
> Where we actually need the warnings are content blocks within the article body and survey/poll description which are image content blocks but don't reference a valid image.

now added warning only for streamfield of articles, and questionnaires description

PR #939, opened 2021-12-21
Remove cache setting

### LINKED ISSUE #819: Remove Settings > Cache Settings
This is obsolete because the "oflline app" banner has been replaced with customizable buttons.

UsmanAli99:
fixes #819

- Remove cache setting


home/models.py
```diff
@@ -731,28 +731,6 @@ class Meta:
         verbose_name_plural = _('Site Settings')
 
 
-@register_setting
-class CacheSettings(BaseSetting):
```
sheralim012, line 735: create migration for this

sheralim012 (changes requested):

geoo89:
It appears that in the second commit, you added the wrong file (cypress DB instead of the generated migrations file).

PR #940, opened 2021-12-22
adds fonts into repository

### LINKED ISSUE #581: Ensure that no resources are required from external domains
As a USER that doesn't have a data package, when I access IoGT for free through my MNO I don't want my experience to be disrupted when external resources don't load.

As a USER I want to be sure that I am not using up my data package if I've been told IoGT is free through my MNO.

Implications on external dependencies (as I understand them, please add)

Critical to address:

- [x] https://fonts.googleapis.com/icon?family=Material+Icons (I assume this is layout-breaking if it doesn't load)

Important to address:

- [ ] Google Analytics (replace with server-side tagging https://developers.google.com/tag-platform/tag-manager/server-side/intro)

Unknown:

- [ ] http://www.w3.org/2000/svg (unsure whether this xmlns tag actually calls the destination URL)

**Urgent tag only applies to Critical items.**

sheralim012:
fixes #581 

- added downloaded Open Sans and Poppins fonts
- removed old imports and updated their usage


geoo89:
From the Free Basics requirements:

"If your service makes use of WOFF webfonts, ensure that the appearance of your site is acceptable if these files are suppressed."
I assume our site looks ok even if these are not loaded?

sheralim012:
we have a list of font families which are used as backups as follow, and at the end browsers sans-serif will be used
```
"Poppins", -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif
```

geoo89:
Thanks. Still looks decent even if we suppress the fonts. I'm happy with this, though I'd prefer an additional approval from someone else. @istride

edmoss94:
image

lgtm, page loads fine even with google fonts blocked

istride (approved):
I think it's ok to merge, but I do have some questions and comments.

Why are the names of the WOFF files like a hash? Would it be better to rename them to something more human-readable?

I think it would be better to define the fonts in a separate CSS file rather than in `iogt.css` and import into `iogt.css`. Better still define in a separate CSS file and combine all CSS files into one for deployment. However, I also think that this is work that is probably best done separately.

PR #942, opened 2021-12-22
Use quiz, poll and survey for multiple attempts error message

### LINKED ISSUE #822: When you cannot take a quiz multiple times, the error message mentions a "survey"
Change to quiz. Check for polls as well.

UsmanAli99:
fixes #822

- Edit multiple attempts error message.
- Use quiz, poll and survey for multiple attempts error message.

### DIFF
--- a/questionnaires/templates/questionnaires/tags/already_completed.html
+++ b/questionnaires/templates/questionnaires/tags/already_completed.html
@@ -1,12 +1,12 @@
 {% load i18n %}
 
 {# use page type when we have translations for poll and quiz #}
-    <p class="survey-page__already-completed">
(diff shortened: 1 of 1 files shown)


geoo89:
Because of the way we do translations, it'd be better to have an if block based on the page type that either renders

{% translate 'You have already completed this survey.' %}
or
{% translate 'You have already completed this poll.' %}
or 
{% translate 'You have already completed this quiz.' %}

geoo89:
Thanks! I'll leave this PR open for now, as I will have to update the PO files and list of strings for translation manager config.
I will do that once a few other tickets that relate to this are addressed.

PR #945, opened 2021-12-23
adds locale detail model

### LINKED ISSUE #798: Option to hide locales from display in language selector
In the Locale settings, a checkbox (default: checked) labeled "Show this locale in the language selector".

If it's unchecked, the language will not appear in the selector but will be accessible at /[XX]/

### LINKED ISSUE #799: Configurable default language
Wagtail has its own notion of default language, but what we need here is the language that is shown when the user visits the page for the first time. For example, if the default language is English, a user visiting central.iogt.site is redirected to central.iogt.site/en/

In addition with #798 (which allows us to hide English from the page altogether if needed), these are the two behaviors we need to make another language than English the default language, without changing Wagtails' default language (which might have other unwanted implications).

sheralim012:
fixes #798
fixes #799

- added LocaleDetail model with one-to-one relation with wagtail Locale model with following fields
  - is_active
  - is_main_language (there can only be one main language)
- added Locale Detail in settings menu
- only active locales will be shown in language switcher
- if page belongs to inactive locale then switcher wouldn't show the language
- inactive locale page could be accessed by manipulating URL and added locale code
- admin locale wouldn't be affected by user locale
- wrote customer middleware to handle language as follow
  - if language is given in path honor that
  - else if **active** langugae given in request (from browser) honor that
  - else if **active** main language is available honor that
  - else if use primary language (LANGUAGE_CODE)


geoo89:
- When I have an existing page without any locale details set, only English is available from the dropdown list. By default, if there's no locale detail entry, a locale should be considered active.

- Spelling: Locale Detail instead of Local Detail.

> else if langugae given in request (from browser) honor that
else if main language is available honor that
else if use primary language (LANGUAGE_CODE)

Main language is the one set in the locale detail, while primary language is hardcoded to be English?

- I tried to make English not available as a language (set is_active to False), while having French both active and as a main language. However when I load the homepage without any language code in the URL and English as browser language, I still get the English page. English also appears in the dropdown list. (If a locale is not active, I should not be able to access it, except _maybe_ if I enter its explicit URL. In fact, it should probably be excluded from the sitemap too, if this is used to determine the pages to download for offline access). Note that English does not appear if the browser language is set to something else.

geoo89:
Could you also run the schema migrations on the cypress DB?
All cypress tests fail now: https://github.com/sheralim012/iogt/runs/4625141764?check_suite_focus=true
(Only the chatbot related test should fail right now.)

sheralim012:
> * When I have an existing page without any locale details set, only English is available from the dropdown list. By default, if there's no locale detail entry, a locale should be considered active.

if locale detail isn't available then locale will be considered active.

> * Spelling: Locale Detail instead of Local Detail.

fixed type

> 
> > else if langugae given in request (from browser) honor that
> > else if main language is available honor that
> > else if use primary language (LANGUAGE_CODE)
> 
> Main language is the one set in the locale detail, while primary language is hardcoded to be English?

primary language is of Django which is set to English in settings (for reference see setting named `LANGUAGE_CODE`)

> 
> * I tried to make English not available as a language (set is_active to False), while having French both active and as a main language. However when I load the homepage without any language code in the URL and English as browser language, I still get the English page. English also appears in the dropdown list. (If a locale is not active, I should not be able to access it, except _maybe_ if I enter its explicit URL. In fact, it should probably be excluded from the sitemap too, if this is used to determine the pages to download for offline access). Note that English does not appear if the browser language is set to something else.

- fixed main language issue, but let's discuss this further on a quick call

sheralim012:
> Could you also run the schema migrations on the cypress DB? All cypress tests fail now: https://github.com/sheralim012/iogt/runs/4625141764?check_suite_focus=true (Only the chatbot related test should fail right now.)

ran migrations and updated cypress db now only chatbot test are failing (for [reference](https://github.com/sheralim012/iogt/runs/4640197192?check_suite_focus=true))

PR #948, opened 2021-12-23
Translate default submit button texts during migration

### LINKED ISSUE #906: Translate button texts during migration
The register button for registration surveys, and submit_button_text for regular surveys (if it was left empty).
Note: This should be done also for the original version of surveys, not only translated versions, as the main language of a site doesn't have to be English. So there should be 4 places where code has to be modified.

Translations of Register and Submit can be drawn from `iogt_content_migrations/files/registration_survey_translations.csv`.

Also, add a warning if the submit button text is longer than 40 characters (i.e. truncated).

geoo89:
fixes #906

Fix of #910 to only update submit buttons that have the default text (and in that case translate the submit button text).


geoo89:
@sheralim012

sheralim012 (approved):

PR #950, opened 2021-12-27
Account related links in footer

### LINKED ISSUE #682: Fixes for account-related links in Footer
1. Logout button does not appear
2. Account-related links appear at the top of the footer - it would be better for them to be in the middle as in V1, or to be placed by the platform administrator

UsmanAli99:
fixes #682

- Add logout button in footer.


edmoss94:
lgtm

PR #956, opened 2022-01-03
:heavy_plus_sign: Add support for V1 migrated URLs

### LINKED ISSUE #850: Redirect V1 URLs
During migration, assume we create a table that maps V1 URLs to V2 page IDs (see #851).

If somebody clicks a link to a V1 page, they land on our V2 site. If we cannot resolve the URL of the V1 page they are trying to visit, look it up in the map and redirect the user to the corresponding V2 page. (Probably using HTTP code 301?)

ehmadzubair:
Closes #850

### DIFF
--- a/home/models.py
+++ b/home/models.py
@@ -1135,7 +1135,11 @@ class V1PageURLToV2PageMap(models.Model):
     @classmethod
     def get_page_or_none(cls, v1_page_url):
         # See https://github.com/unicef/iogt/issues/850 for more details on why /home/ is prepended
-        obj = cls.objects.filter(v1_page_url=f'/home{v1_page_url}').first()
+        urls_to_match = [
+            f'/home{v1_page_url}',
+            f'/home{v1_page_url}/'
+        ]
+        obj = cls.objects.filter(v1_page_url__in=urls_to_match).first()
         return obj.v2_page if obj else None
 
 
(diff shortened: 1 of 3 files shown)


geoo89:
This doesn't appear to fully work.

Example: https://mz.goodinternet.org/footer-pages/about-internet-good-things/
After MZ migration, and setting PT as main language and disabling English (in the locale detail), I enter http://localhost:8000/footer-pages/about-internet-good-things/ in the browser. I get redirected to http://localhost:8000/pt/footer-pages/about-internet-good-things/ which gives a 404

ehmadzubair:
@geoo89 can we test this on a call?

geoo89:
Sorry, I'm only free in the evenings for the next two days.
But I had a look into the V1PageURLToV2PageMap entries. It looks like there's a `home` prefix in the table that shouldn't be there. Here's an example entry: `/home/footer-pages/about-internet-good-things/ ` --> `/pt/footers/about-internet-good-things/`
I've only checked for MZ for now, but I assume this is common across all sites.

I guess that should be fixed as part of the migration rather than within the middleware? What do you think?

`http://localhost:8000/home/footer-pages/about-internet-good-things/` redirects fine, so I think this component actually works as intended. Only minor thing is that `http://localhost:8000/home/footer-pages/about-internet-good-things` yields a 404, and I think we should make this work as well.

ehmadzubair:
Resolution: Remove `/home` if the url starts with `/home`

geoo89:
Thanks! Could you also please address my other remark?

> Only minor thing is that `http://localhost:8000/footer-pages/about-internet-good-things` (without trailing `/`) yields a 404, and I think we should make this work as well.

PR #959, opened 2022-01-04
Serve static assets via separate http server

istride:


docker-compose.yml
```diff
@@ -1,19 +1,17 @@
 version: '3.5'
 
 services:
-
   django:
     build:
       context: ./
-      dockerfile: Dockerfile.dev
-    container_name: dc01
+      dockerfile: Dockerfile
```
ehmadzubair, line 7: I think we should make a `docker-compose.prod.yml` for this?

home/templatetags/generic_components.py
```diff
@@ -9,24 +9,28 @@
...
+        return {
+            'title': title,
+            'href': href,
+            'icon_path': icon_path,
+            'font_color': font_color,
+            'background_color': background_color,
+            'extra_classnames': extra_classnames,
+        }
+    except:
+        return {}
```
ehmadzubair, line 33: I think this except block is too generic. This will catch all the exceptions and we probably don't want that

home/templatetags/image_tags.py
```diff
@@ -1,22 +1,34 @@
...
 @register.simple_tag
 def svg_to_png_url(svg_relative_path, fill_color=None, stroke_color=None,):
-    absolute_path = f'{settings.BASE_DIR}{svg_relative_path}'
-    return SVGToPNGMap.get_png_image(absolute_path, fill_color=fill_color, stroke_color=stroke_color).url
\ No newline at end of file
+    absolute_path = Path(settings.BASE_DIR) / svg_relative_path
+    try:
+        return SVGToPNGMap.get_png_image(absolute_path, fill_color=fill_color, stroke_color=stroke_color).url
+    except:
+        message = f"Failed to convert SVG to PNG, file={absolute_path}"
```
ehmadzubair, line 32: This is a generic except block. Should we make it more specific?

iogt/settings/production.py
```diff
@@ -1,7 +1,7 @@
 import os
 from .base import *
 
-DEBUG = True
+DEBUG = os.environ.get('DEBUG', 'False') == 'True'
```
ehmadzubair, line 4: Should we have this as a boolean? bool('False') evaluates to True

ehmadzubair, line 4: Should we use `bool(int(os.environ.get('DEBUG', 0))`

home/templatetags/generic_components.py
istride: I admit that this error handling needs some more work. My rationale for this is that we would not want a missing SVG file, for example, to cause an internal server error. I would rather catch any (or most) errors, log it and move on. If that means there would be a missing button on the front-end, so be it.

iogt/settings/production.py
istride: Why?

In any case, I expect this to be a temporary option as DEBUG should never be 'True' in production.

PR #962, opened 2022-01-05
:wrench: Update create_default_site command

### LINKED ISSUE #836: Posting comment redirects to example.com
For instance, here: https://global.iogt.site/fr/sections/emergency-information/about-covid-19/about-coronavirus/

Site object is configured as global.iogt.site port 80. Also tried global.iogt.site:443, http://global.iogt.site:80, https://global.iogt.site:443, same result regardless.

ehmadzubair:
Closes #836

### DIFF
--- a/home/management/commands/create_default_site.py
+++ b/home/management/commands/create_default_site.py
@@ -8,9 +8,17 @@ User = get_user_model()
 
 class Command(BaseCommand):
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--domain',
+            help='The domain on which the website is deployed'
+        )
+
     def handle(self, *args, **options):
+        domain = options.get('domain') or 'localhost:8000'
+
         _, created = Site.objects.update_or_create(pk=settings.SITE_ID, defaults={
-            'domain': 'localhost:8000',
+            'domain': domain,
             'name': 'default'
         })
         self.stdout.write(self.style.SUCCESS(f'Site object {"created" if created else "updated"}'))
--- a/home/management/commands/create_default_site.py
+++ b/home/management/commands/create_default_site.py
@@ -1,7 +1,10 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.contrib.sites.models import Site
+from django.contrib.sites.models import Site as DjangoSite
 from django.core.management.base import BaseCommand
+from wagtail.core.models import Site as WagtailSite
+
+from home.models import HomePage
 
 User = get_user_model()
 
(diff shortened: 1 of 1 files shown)


geoo89:
Could you please make this command update both the Site object in django (which is currently does) but also the wagtail Site object which we see in the CMS? Thanks!

sheralim012:
@geoo89 done

home/management/commands/create_default_site.py
```diff
@@ -1,16 +1,33 @@
...
+        domain = options.get('domain') or 'localhost:8000'
+
+        __, created = DjangoSite.objects.update_or_create(pk=settings.SITE_ID, defaults={
+            'domain': domain,
             'name': 'default'
         })
-        self.stdout.write(self.style.SUCCESS(f'Site object {"created" if created else "updated"}'))
+        self.stdout.write(self.style.SUCCESS(f'Django Site object {"created" if created else "updated"}'))
+        __, created = WagtailSite.objects.update_or_create(is_default_site=True, defaults={
+            'hostname': domain,
```
geoo89, line 29: Only the hostname (and port) should be updated, not the other fields of the site object.

home/management/commands/create_default_site.py
```diff
@@ -1,16 +1,33 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--domain',
+            help='The domain on which the website is deployed'
+        )
+
     def handle(self, *args, **options):
-        _, created = Site.objects.update_or_create(pk=settings.SITE_ID, defaults={
-            'domain': 'localhost:8000',
+        domain = options.get('domain') or 'localhost:8000'
```
geoo89, line 21: It appears that the DjangoSite domain does not contain a port (at least, if I specify a port, I get an error when posting a comment, see #836, while it works ok if I only specify the URL), so the default should probably be simply `localhost`?

home/management/commands/create_default_site.py
```diff
@@ -1,16 +1,33 @@
...
 User = get_user_model()
 
 
 class Command(BaseCommand):
 
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--domain',
+            help='The domain on which the website is deployed'
+        )
```
geoo89, line 18: Can we add a port parameter which affects the WagtailSite object? (It seems that the DjangoSite object does not have a port associated with it)

geoo89 (commented):
A few issues to address here.

sheralim012:
@geoo89 added port argument for wagtail site object and simple domain argument will be used for django domain/hostname

PR #974, opened 2022-01-07
create locale details in content migration

### LINKED ISSUE #968: Create locale detail during migrations
For most sites, this will not be needed, but if a language different from English is the main language, we should have a corresponding locale detail entry. Similarly, if English is not active (e.g. mz.goodinternet.org) then there should be an entry disabling English.

sheralim012:
fixes #968 

- create locale details in content migration
- all locale details will be active by default
- main language will be set from v1 db


sheralim012:
@geoo89 changes addressed

PR #993, opened 2022-01-13
fixed live/deleted content rendering

### LINKED ISSUE #985: When features are embedded in a page and then unpublished, the page where they are embedded should load without the embedded content
Currently, if a page is created with an embedded poll and then published, but then the poll is later unpublished, the page where the poll is embedded will not load (front end users receive an error syntax page). Ideally, in this scenario, the page will still load, but without the unpublished poll and the CMS editor will not be able to republish or resave the page without deleting the poll widget component from page content.

sheralim012:
fixes #985 

- homepage
  - banners: have valid page, must be live
  - featured content
    - article: have valid page, must be live
    - page button: have valid page, must be live
    - embedded questionnaire: have valid page, must be live
- section
  - page link page: have valid page, must be live
  - section: must be live
  - article: must be live
  - poll: must be live both in direct and card mode
  - survey: must be live both in direct and card mode
  - quiz: must be live both in direct and card mode
  - top level sections: must be live
- article
  - poll: must be live both in direct and card mode
  - survey: must be live both in direct and card mode
  - quiz: must be live both in direct and card mode
  - page button: have valid page, must be live
  - recommended articles: have valid page, must be live
- footers
  - page link page: have valid page, must be live
  - others: must be live
- questionnaires
  - page button: have valid page, must be live

### DIFF
--- a/home/blocks.py
+++ b/home/blocks.py
@@ -92,10 +92,12 @@ class EmbeddedPollBlock(EmbeddedQuestionnaireBlock):
 
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context)
-        context.update({
-            'direct_display': value['direct_display'],
-            'page': value['poll'].specific,
-        })
+        poll = value.get('poll')
+        if poll and poll.live:
+            context.update({
+                'direct_display': value['direct_display'],
+                'questionnaire': poll.specific,
+            })
         return context
 
     class Meta:
@@ -107,10 +109,12 @@ class EmbeddedSurveyBlock(EmbeddedQuestionnaireBlock):
 
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context)
-        context.update({
-            'direct_display': value['direct_display'],
-            'page': value['survey'].specific,
-        })
+        survey = value.get('survey')
+        if survey and survey.live:
+            context.update({
+                'direct_display': value['direct_display'],
+                'questionnaire': survey.specific,
+            })
         return context
 
     class Meta:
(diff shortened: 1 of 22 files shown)


geoo89:
Thanks, this seems pretty thorough. I could not create any crashes/404s on the user-side by deleting/unpublishing pages.

A few minor things though:
- The section text in the article card lost its color, e.g. Covid-19 should be red, could you bring that back?
image

Only fix the issues below if they are very easy, otherwise I will create separate issues for them:
- If a hyperlink in a paragraph references a draft page, it will still be rendered as a link --> 404 if you click it. (For deleted pages, it renders as plain text, which is good)
- If a PLP in a footer refers to a draft page, it still displays the footer button (pressing it redirects to the page you're currently on.)
- If I try to delete an image (that is used as lead image/banner image/maybe some more) in the CMS, or if I try to delete a page that is referenced by PLP, the page crashes with `ProtectedError at /admin/images/9/delete/ ("Cannot delete some instances of model 'Image' because they are referenced through protected foreign keys...`. Would be nice to handle this more gracefully, like for chatbot channels or SVGs (for svg, it shows the text: 'busy-24-hour' is currently referenced by other objects, and cannot be deleted without jeopardising data integrity. To delete it successfully, first remove references from the following objects, then try again)

sheralim012:
@geoo89 I have fixed the article card font color issue, but other issues aren't quite straightforward so could you create separate tickets for these

geoo89:
Thank you. I created separate issues. #1001, #1002, #1003, #1004, #1005

PR #1007, opened 2022-01-18
adds accessibility text for quiz answers

### LINKED ISSUE #338: Highlighting correct and wrong answers in quizzes
Current scheme:
- Green tick: You picked and it was correct. 
- Red cross: You picked and it was wrong, OR you didn't pick but it was correct.
- Grey: You didn't pick it and it was incorrect
image
Thus from the results page you cannot infer which answers are correct without knowing what you picked. 

I assume the emphasis should be on what the correct answers are, rather than on what you did wrong.

Proposal (for discussion):
Use the checkbox to indicate what was picked and color to indicate the correct answers:
- Green tick: correct and you picked it
- Green empty: correct but you missed it
- Red cross: incorrect but you picked it
- Grey: incorrect and you didn't pick it

sheralim012:
fixes #338 

- used [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) for accessibility as span tag doesn't have any alt text option

### DIFF
--- a/questionnaires/templates/quizzes/quiz_landing.html
+++ b/questionnaires/templates/quizzes/quiz_landing.html
@@ -104,11 +104,11 @@
                             {% if field|length > 1 %}
                                 {% for field_option in field %}
                                     <div class="quest-item__content" tabindex="0">
-                                        <label class="cust-check cust-check--2 quest-item__input
-                                                {% get_answer_rendering_class field field_option fields_info %}"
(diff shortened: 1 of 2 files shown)


cbunicef:
Confirming that unicef.org also uses aria-label for alt text when role="img":

    <span class="b-image" role="img" aria-label="Girl smiling holding a UNICEF book">

PR #1024, opened 2022-01-20
adds logic for optional registration survey

### LINKED ISSUE #896: Icons are not accessible when a user is logged in but has not completed required registration
Before registration completed

image

After registration completed

image

### LINKED ISSUE #944: Less strict registration survey
On the MZ site, the registration survey consists of the optional field "Date of birth". In V1, it is presented on a separate page, no submission of that data is needed to access IoGT. In V2, the form has to be submitted (even if the field is left blank). This is too strict. Note: On most other V1 sites, registration is a one-step process where the registration survey questions are displayed on the same page as the name/password field.

Two possible solutions for V2:

- One-step process: Show the registration survey embedded on the same page as the other registration forms (name, displayname, password).
- Two-step process: Keep everything as it is, with two exceptions:
    - Users are initially directed to, but do not have to submit the registration survey if it only contains optional questions (i.e. if you are in the registration survey, and you click a link to some other page, you will NOT be redirected to the survey, but are free to browse the site)
    - If the registration survey contains required questions, and the user clicks a link to some other page, they do get redirected to the survey (as currently happens), but additional red text appears stating "Please complete the questions marked as required to continue"

Also see https://github.com/unicef/iogt/issues/899#issuecomment-994780752

sheralim012:
fixes #944 
fixes #896

- registration survey with all optional fields could be skipped 
- show a message to fill required fields for the registration survey


iogt_users/middlewares.py
```diff
@@ -29,12 +30,29 @@ def __call__(self, request):
...
+        if (is_registered_user
+                and not user.has_filled_registration_survey
+                and not is_url_allowed
+                and registration_survey
+                and registration_survey.has_required_fields()):
+            should_redirect_to_registration_survey = True
+            messages.add_message(
+                request, messages.ERROR, 'Please complete the questions marked as required to continue')
+
+        if (is_registered_user
```
ehmadzubair, line 46: else if

geoo89:
When there is a required question, and I click a link to leave the questionnaire, I get redirection to the questionnaire (intended), but also the warning appears twice now:

image

Ideally, don't show the warning the first time around, and only show it once the user tries to leave the registration survey.


Also, technically, the registration survey in some language could have some required fields while in another it doesn't. This currently doesn't work, but I guess it's an unintended edge case anyway so leave it unless it's trivial to fix.

iogt_users/middlewares.py
```diff
@@ -29,12 +30,29 @@ def __call__(self, request):
...
+
+        should_redirect_to_registration_survey = False
+        if (is_registered_user
+                and not user.has_filled_registration_survey
+                and not is_url_allowed
+                and registration_survey
+                and registration_survey.has_required_fields()):
+            should_redirect_to_registration_survey = True
+            messages.add_message(
+                request, messages.ERROR, 'Please complete the questions marked as required to continue')
```
geoo89, line 44: Add translate tag

sheralim012: done

sheralim012:
* don't show warning for the first time
* use localized registration survey to check required fields

unable to reproduce double warning, let's sync on a quick call

sheralim012:
fixed images not loading for footers in registration survey

PR #1036, opened 2022-01-21
single column menu button

### LINKED ISSUE #1029: Add method for navigating to Menu (ie section list) on mobile
As a MOBILE USER I want to be able to easily get to the navigation menu on the website. 

Currently the navigation is below all content, including on the homepage, and we don't have support for a Flat Menu button that either 1) expands from the Flat Menu to list top-level sections 2) clicks to a list of top-level sections.

The simplest possible solution may be adding an HTML anchor at the top of the navigation menu, and using a Flat Menu button to link to that anchor. This doesn't work well on desktop, but solves the problem on mobile platforms.

sheralim012:
fixes #1029 

- adds "Display only in single column view" flat menu item field
  - menu items with this enabled will show only in single column view
- Link to customer URL with #menu will jump user to footer in single column view
- added data migration to create menu and flat menu for above functionality for all available locales
Screenshot 2022-01-21 at 6 59 10 PM


home/migrations/0040_auto_20220121_1337.py
```diff
@@ -0,0 +1,34 @@
...
+    locales = Locale.objects.all()
+    for locale in locales:
+        menu, __ = FlatMenu.objects.update_or_create(handle=f'{locale.language_code}_menu_live', defaults={
+            'title': f'{locale.language_code} main menu',
+            'site': site,
+            'heading': 'Main Menu',
+        })
+        IogtFlatMenuItem.objects.update_or_create(link_url='#menu', menu=menu, defaults={
+            'link_text': 'Menu',
+            'display_only_in_single_column_view': True
```
geoo89, line 22: Would it be possible to also add the `iogt/static/icons/burger.svg` icon here?

home/migrations/0040_auto_20220121_1337.py
```diff
@@ -0,0 +1,34 @@
...
+    Locale = apps.get_model('wagtailcore.Locale')
+    Site = apps.get_model('wagtailcore.Site')
+    FlatMenu = apps.get_model('wagtailmenus.FlatMenu')
+    IogtFlatMenuItem = apps.get_model('home.IogtFlatMenuItem')
+
+    site = Site.objects.get(is_default_site=True)
+    locales = Locale.objects.all()
+    for locale in locales:
+        menu, __ = FlatMenu.objects.update_or_create(handle=f'{locale.language_code}_menu_live', defaults={
+            'title': f'{locale.language_code} main menu',
```
geoo89, line 16: It'd be better not to overwrite these values if a FlatMenu already exists. Use `get_or_create`?

home/migrations/0040_auto_20220121_1337.py
sheralim012: done

home/migrations/0040_auto_20220121_1337.py
sheralim012: done

PR #1040, opened 2022-01-21
fixes size for images of stream block

### LINKED ISSUE #1038: Image sizes within paragraphs
We can insert images into articles/questionnaires via image stream blocks or by adding them to a paragraph.

Consider serving these images also at 360 px width quality. (I believe they are currently served at original size.)

sheralim012:
fixes #1038 

- fixes size for images in streamblock


home/templates/blocks/image.html
```diff
@@ -0,0 +1,3 @@
+{% load wagtailimages_tags %}
+
+{% image value width-320 %}
```
geoo89, line 3: This should be 360 I think

geoo89 (commented):
Would it be complicated to also add this for images that are contained within paragraphs?

geoo89:
> Would it be complicated to also add this for images that are contained within paragraphs?

Just to confirm, we'd ideally like the serving size of 360 px width also for images within paragraphs.

home/templates/blocks/image.html
sheralim012: done

sheralim012:
fixed size for images in paragraphs as 360

home/migrations/0042_auto_20220125_1128.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def _fix_article_paragraph_image_format(apps, schema_editor):
+    Article = apps.get_model('home', 'Article')
+    format_regex = re.compile('format="(left|right)"')
+    articles = Article.objects.all()
+    for article in articles:
+        for block in article.body.stream_data:
+            if block.get('type') == 'paragraph':
+                block['value'] = format_regex.sub('format="fullwidth"', block.get('value'))
```
geoo89, line 14: Am I correct in that you're overwriting the information on whether an image should be left- or right- aligned here by making all images full-width? I'm concerned about this as this sounds like information we might want to retain, especially for smaller images (which would now get stretched across the whole screen). We might want to serve images in a more flexible way in the future, and I don't think we should throw this information away.

Could we simply add the 360 px width tag to the left/right aligned image as well, even if that means on desktop some left/right aligned images will not show at full width anymore as they are served at a smaller quality than the column width?

geoo89 (commented):
All images are served at the desired quality now, however I have a concern about throwing away the left/right alignment information of images (if I understand correctly that this is what is happening). @cbunicef What is your opinion on this?

home/migrations/0042_auto_20220125_1128.py
geoo89: Also @cbunicef

cbunicef: Yes, we shouldn't discard the right/left align data during migration.

cbunicef: We shouldn't throw away the left/right alignment data. I believe our intended solution here is on the platform codebase, rather than the migration.

If we can make the paragraph block interpret images as `{% image [image] width-360 %}` then I think we end up with an image that will be MAX 360 wide, and will also only show as normally expected.

60x60 image is served as 60x60 and keeps its alignment or full-width
700x700 image is served as 360x360 and keeps its alignment or full-width
Both are resized in browser to be appropriate size based on alignment/full-width and column size.

sheralim012: - we can't serve images from paragraph block as `{% image [image] width-360 %}` because they are stored as tags in the database
- changing format (in our case image size) is the recommended way from wagtail

image sizes 
| label       | default | proposed |
|-------------|---------|----------|
| full width  | 800px   | 360px    |
| left align  | 500px   | 260px    |
| right align | 500px   | 260px    |

changes default sizes will only require codebase change, and we wouldn't have to perform any kind of data/schema migration

cbunicef: What about if we use 
full width = width-360
left/right align = max-360x5000

sheralim012: we first unregister those then register new ones (with proposed image sizes) with the same name

PR #1051, opened 2022-01-24
minor fixes for rendering

sheralim012:
- use class method to check first content instead of template tag
- use page variable instead of self in template
- removes undefined navbar background color reference
- update redirection logic to skip jsi18n
- updates questionnaires terms and conditions templates


geoo89:
Is this PR still relevant? If the changes in it add additional safety guards, it might still be worthwhile cleaning up this PR and merging it?

sheralim012:
once #1062 is merged I'll see if any work is required on this PR

PR #1058, opened 2022-01-25
Translation tracking process

edmoss94:
Taking draft translation code written by Georg and combining together into a management function 'translation_tracking'. Function updates the excel translation files then updates the PO files accordingly


common/translation_utils/update_status.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def needs_improvement(self):
+        return self.bad_source_phrase != ''
+
+    def is_phrase(row):
+        return row[0] in ['', 'sg', 'pl']
+
+def update_status_run():
+    
+    def lookup_status(msg, replacement_strings=[]):
```
geoo89, line 20: It's pretty uncommon to define functions within other functions.
You can just keep them outside of `updated_status_run`.

common/translation_utils/make_pos.py
```diff
@@ -0,0 +1,136 @@
...
+import csv
+import copy
+import os
+
+class Translation(object):
+    pass
+
+def make_pos_run():
+
+    def make_translation():
```
geoo89, line 11: It's pretty uncommon to define functions within other functions.
You can just keep them outside of `make_pos_run`.

common/translation_utils/make_pos.py
```diff
@@ -0,0 +1,136 @@
...
+            translation = Translation()
+            po = polib.POFile()
+            po.metadata = copy.deepcopy(po_metadata_template)
+            translation.pofile = po
+            return translation
+
+    po_metadata_template = {
```
geoo89, line 18: This is also used in `make_testing_pos`. Maybe factor it out into a separate file and create a `get_po_metadata_template` function that returns a template. Then import and use that function for both of these scripts.

(Note: The reason to wrap it into a function is that you get a new instance of the template each time, rather than sharing the same instance across multiple scripts.)

home/management/commands/translation_tracking.py
```diff
@@ -0,0 +1,29 @@
+from django.core.management.base import BaseCommand
+from django.core.management import call_command
+from common.translation_utils.update_status import *
```
geoo89, line 3: `from ... import *` is generally advised against as it pollutes the namespace.
Only import what you need, e.g. `from common.translation_utils.update_status import `update_status_run`

home/management/commands/translation_tracking.py
```diff
@@ -0,0 +1,29 @@
+from django.core.management.base import BaseCommand
```
geoo89, line 1: Would it make more sense to move this file to `common/management/commands/translation_tracking.py`? Does that still work?

common/translation_utils/make_pos.py
```diff
@@ -0,0 +1,136 @@
...
+    for translation in translations:
+        os.makedirs(translation.path, exist_ok=True)
+        filename = os.path.join(translation.path, "django.po")
+        translation.pofile.save(filename)
+    for translation in translationsjs:
+        filename = os.path.join(translation.path, "djangojs.po")
+        translation.pofile.save(filename)
+
+    with open("home/translatable_strings.py", 'w') as f:
+        f.write('translatable_strings = ' + str(translatable_strings))
```
geoo89, line 136: Nicer formatting:

```
content = "\n    ".join(e.__repr__() for e in translatable_strings)
f.write(f'translatable_strings = [\n    {content}\n]')
```

geoo89 (commented):
Just a few style comments for now.

Will look at the functionality in more detail and we can discuss later.

geoo89:
I made a couple of fixes myself already. You can merge in https://github.com/geoo89/iogt/tree/translations to continue from there.

edmoss94:
Have addressed majority of style comments addressed previously

common/translation_utils/update_status.py
edmoss94: Have removed nested functions

common/translation_utils/make_pos.py
edmoss94: Have removed nested functions

common/translation_utils/make_pos.py
edmoss94: Have factored out into a separate file as advised

home/management/commands/translation_tracking.py
edmoss94: have changed to reference specific functions

common/translation_utils/make_pos.py
edmoss94: Has been addressed

home/management/commands/translation_tracking.py
edmoss94: Have tried this and it didn't work, but not really sure why, will read up a bit more on django to understand the management  function requirements better

PR #1069, opened 2022-01-27
removes footer page

### LINKED ISSUE #1082: Remove footer page type
To ensure no footers are in use, we added code that makes it impossible to add new footers (#1070).
Once this is deployed on all sites, we can check that no footer pages have been added (and remove them if necessary) and then remove the footer type from the codebase.

sheralim012:
fixes #1082

- removed footer page
- updated content migration to remove footer page reference
- updated create initial data command to use article as footer
- updated sitemap to remove footer page reference


sheralim012:
outdated, will create new PR if needed.

PR #1105, opened 2022-02-02
filter out registration survey from form data listing

### LINKED ISSUE #1102: Registration Surveys should not appear in Form Data tab
The functions for viewing survey results and accessing all contact information should be separated, so that permissions are easily manageable. "View user" permission in Group settings should be the only way to view user info.

This should close #1097

sheralim012:
Closes #1102 

- filter out registration surveys from "Form Data" questionnaires listing


sheralim012:
@geoo89 I have disabled the submissions panel for registration surveys

PR #1106, opened 2022-02-02
updates content migration to migrate v1-list as v2-markdown

### LINKED ISSUE #1087: During migration, move List blocks to lists in Paragraph blocks
List blocks in V1 are a compact list of single-row fields for markdown entry.

image

List blocks in V2 are a very long list of large fields for markdown entry.

image

It would be better to migrate List blocks to Paragraph blocks which are easier to manage:

image

sheralim012:
Closes #1087 

- updated content migration to migrate v1-list as v2-markdown with list

p.s: we aren't fixing the existing data (deployed sites) as of now


geoo89:
I noticed that the line spacing is different between list elements and lists in markdown blocks:

image

I assume this difference is acceptable?

Issue #514, opened 2021-08-31
Offline Access limitations


ehmadzubair:
Here's a list off the issues related to Offline access and their possible fixes.

| Limitation                                          | Fixes                                                                                                                                |
|-----------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| Polls, Surveys and Quizzes can’t be submitted       | Attempt to make multi-step forms appear as single page forms, Store the submissions locally until there is an internet connection |
| User article “read” functionality is not accessible | Store the used article “read” activity until there is an internet connection                                                         |
| User read progress isn’t updated                    | Disable user progress bar in offline mode                                                                                            |
| User Login/Logout, update profile is not available  | Hide Login/Logout/My Profile buttons in offline mode                                                                                 |
| Search doesn’t work                                 | Hide the search bar in offline mode                                                                                                  |
| The external link intermediate page isn’t cached    | Update sitemap script to include all external links to fix the broken page                                                           |

@sheralim012 @geoo89 @volloholic

volloholic:
Creating each point as a task and responding to each point.
- [ ] Polls, Surveys and Quizzes offline: Looks good but there are still lots of details to dig into (i.e. I assume polls and quizzes will not be able to have the normal completion pages what do we have instead?)
- [ ] Store User article “read” info: Sounds good but we should almost certainly store the full read info, see next point.
- [ ] User read progress isn’t updated: If we are storing the full "read" activity offline then we should be able to do this through javascript.
- [ ] Hide Login/Logout/My Profile buttons in offline mode: Fine for now
- [ ] Hide the search bar in offline mode: Fine for now
- [ ] Update sitemap script to include all external links to fix the broken page: This doesn't sound right to me but maybe I'm not understanding? Should not include external pages but should include a page which says you need internet and this may use data. This probably needs discussion.

geoo89:
A few detail that additionally need to be thought out:

- The external link site should maybe read something slightly different, e.g. that you'll need internet access to proceed to the next page?
- log-in only/single-submission questionnaires: hide the former from the user? Only accept the first submission for the latter?
- questionnaires landing page: A placeholder indicating that your responses have been saved and will be uploaded once you come online? For polls the old results could possibly displayed?
- Any chatbot buttons/functionality should be disabled

cbunicef:
For multi step -> single Page quizzes/surveys. Unfortunately we can't just do this simple fix because survey design can be very specific.

[...]

PR #1126, opened 2022-02-07
show/hide UI components for offline access

### LINKED ISSUE #881: Offline access (minimal)
Simplification of #512 so we can have this somewhat presentable sooner:

As before, disable:
- Search
- commenting
- password change
- chatbot buttons
- offline access button
- (any other kind of form submissions)

For now also disable:
- questionnaire submissions
    - Display a text that you cannot submit surveys when offline
    - Disable submit/next button
- reading progress

sheralim012:
Closes #881 

- show/hide following UI components in online/offline mode
  - search input and search button
  - comments
    - posting
    - liking comment
    - replying comment
  - chatbot buttons
  - offline app buttons
  - accounts
    - login
    - logout
    - create account
    - change digital pin
  - progressbar
- for questionnaires submit/next button
  - append text "(You cannot submit when offline)"
  - disable the button


ehmadzubair (approved):

ehmadzubair (approved):

geoo89:
For reading progress, the articles that have been read are still being highlighted: 
image

For the offline button, could we grey it out + make it unclickable and render it with the "offline" message that is specified when you create such a button?

Similarly, could we grey out the chatbot button rather than hiding it altogether?

Report button still appears for comments.

Questionnaire submit buttons still seem to appear as normal -- did I resolve the merge conflicts incorrectly?

Logout button appears on the profile page. Maybe this needs discussion and we can defer it to later, also see https://github.com/unicef/iogt/issues/514#issuecomment-1020297835 
When the site is cached, are the pages loaded as if the user was logged in (when they are logged in)?

sheralim012:
> The lock is now missing from the "change password" button when I'm online.

could you try `python manage.py collectstatic` 

> The chatbot button now redirects me when I'm offline, but doesn't do anything when I'm online. This should be reversed.

somehow it was removed in the last commit, added back the handlers for it

> The report button now does not appear at all, also when online.

fixed

> Could you merge/rebase develop again, now that it has passing tests, to check whether this PR makes any tests fail?

done

PR #1134, opened 2022-02-08
fixes #1018 allows banners to not contain links

### LINKED ISSUE #1018: Decorative banners
There are use cases for banners that don't link to a page, e.g. as visual separators.

To make this intent explicit, maybe add a checkbox to be ticked for a banner to render without a link,
so that if the referenced page is invalid/blank (e.g. due to some page deletion) it will not accidentally render.

ChrisMarsh82:
updated HTML so link would not be added if it didn't exist.
Added logic to add banners if linked page is published or if there is no link


ChrisMarsh82:
@ehmadzubair @sheralim012 - can one of you check this is the right way to do the HTML?

home/templates/home/tags/banners_list.html
```diff
@@ -2,8 +2,12 @@
 
 <section class='banner-holder'>
     {% for banner in banners %}
-        <a href="{% pageurl banner.banner_link_page %}">
-            {% image banner.banner_image width-360 %}
-        </a>
+        {% if banner.banner_link_page == None %}
```
sheralim012, line 5: `{% if banner.banner_link_page %}` is sufficient

home/templates/home/tags/banners_list.html
```diff
@@ -2,8 +2,12 @@
...
+            <a>{% image banner.banner_image width-360 %}</a>
```
sheralim012, line 6: could you try using span instead of a tag as it shouldn't be clickable without linked page

sheralim012 (changes requested):

geoo89:
Obsolete due to #1239

PR #1186, opened 2022-03-03
adds server-side gtm container

### LINKED ISSUE #581: Ensure that no resources are required from external domains

sheralim012:
Closes #581 

- added server-side container in docker compose
- updated client-side gtm configs

### DIFF
--- a/iogt/templates/ga_tag_manager.html
+++ b/iogt/templates/ga_tag_manager.html
@@ -1,0 +1,21 @@
+{% if settings.home.SiteSettings.local_ga_tag_manager %}
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ settings.home.SiteSettings.local_ga_tag_manager }}"
+    height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+
+    <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+    new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+    j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+    'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+    })(window,document,'script','dataLayer','{{ settings.home.SiteSettings.local_ga_tag_manager }}');</script>
+{% endif %}
+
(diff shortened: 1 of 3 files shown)


iogt/templates/ga_tag_manager.html
```diff
@@ -0,0 +1,21 @@
+{% if settings.home.SiteSettings.local_ga_tag_manager %}
+    <noscript><iframe src="https://www.googletagmanager.com/ns.html?id={{ settings.home.SiteSettings.local_ga_tag_manager }}"
```
istride, line 2: This would trigger a request to an external domain. Hopefully, 'gtm.js' can be self-hosted, but what can we do about this iframe?

istride (commented):
What's the difference between 'local' and 'global' ids for tag manager and analytics?

Is the intention here to host our own GTM server and make it available at, say, 'www.internetofgoodthings.org/analytics/'?

iogt/templates/ga_tag_manager.html
sheralim012: could you check these articles
- https://developers.google.com/tag-platform/tag-manager/server-side/custom-domain
- https://stape.io/blog/how-to-add-a-custom-domain-to-the-google-tag-manager-server-container

these explain how to configure custom domain for google tag manager

istride: I'm finding it difficult to see how those two links address my question. My understanding is that we cannot have references to, for example, 'www.googletagmanager.com', so surely more work needs to be done here to make those references configurable.

sheralim012: my understanding is that if we can configure custom domains (IoGT domains) as mentioned in the above articles then `gtm.js` and other assets could be served from those domains,
and making those base URLs configurable wouldn't take long, we can put those in site settings too

PR #1189, opened 2022-03-08
adds wagtail transfer lookup fields

### LINKED ISSUE #1183: Import tool failing
When testing the import tool with this set-up (migrating a small sub-section into the Rwanda site): 
image
We receive this error message:
image

sheralim012:
Closes #1183 

- added `username` in wagtail transfer lookup fields


sheralim012:
@istride could you review

cbunicef:
From my limited understanding of WAGTAILTRANSFER_LOOKUP_FIELDS, is this enough to identify the unique Page across multiple platforms?

slug + language code + username (presumably original author)

There will definitely be some instances of multiple Pages with the same slug across platforms. It doesn't seem feasible to prevent and I don't think we would want to.

PR #1193, opened 2022-03-08
optimize site and reduce page load

### LINKED ISSUE #1184: Speed up initial load
Currently, the web page loads slower than V1. see https://www.webpagetest.org/video/compare.php?tests=220223_AiDc5G_DDD,220223_BiDcMX_CY6

sheralim012:
Closes #1184 

- optimized fetching banner pages with linked pages
- optimized color scheme fetching for buttons, sections, articles, questionnaires
- optimized fetching of SVG to PNG map
- optimized fetching of translations
- added conn max age environment variable for postgres
- added cached session backend
- added switchable cache backend (db/redis/default)

if you want to use database cache then update `docker-compose.yml` to set 'db' as CACHE_BACKEND and run the following command to create database cache table
```
python manage.py createcachetable
```
if you want to use redis cache then update `docker-compose.yml` and set 'redis' as CACHE_BACKEND


iogt/settings/base.py
```diff
@@ -423,8 +424,41 @@
...
+            "OPTIONS": {
+                "CLIENT_CLASS": "django_redis.client.DefaultClient",
+                "IGNORE_EXCEPTIONS": True,
+            },
+        },
+        'renditions': {
+            'BACKEND': 'django_redis.cache.RedisCache',
+            'LOCATION': 'redis://redis:6379/1',
+        },
+    }
```
ehmadzubair, line 464: raise exception in else

sheralim012: done

iogt/middleware.py
```diff
@@ -96,3 +99,35 @@ def process_response(self, request, response):
                 return HttpResponsePermanentRedirect(page.url)
 
         return return_value
+
+
+class GlobalContextMiddleware:
```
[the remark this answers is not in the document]
sheralim012: done

geoo89:
I tested this in dev.

- translations added in the translation manager never got refreshed.
- Page preview is not always instant:
    1. Edit a page
    2. Preview (the changes are visible)
    3. Edit the page again
    4. Preview again (the new changes are not visible and take a while to propagate)
- When a user posts a comment, and they are redirected to the current page, their comment is not shown until the next cache refresh. This is confusing.
- Same for responding to the chatbot.

Page edits took a while to propagate to the live pages. I assume this is intentional. Some changes however cannot be previewed, e.g.
- Changing the color of a top-level section
- Editing/Adding a nav-bar entry
- Adding/editing footers
As not all pages are refreshed simultaneously, some pages show the changes earlier than others.
Note that global themeing does immediately take effect.

It seems like all but the translation issue are related to the page-level caching. Should we disable this for now and see how much speedup we get without it?

ChrisMarsh82:
Split this into 2 PRs. Separating caching from other changes.
Add option to have Redis, DB, or no cache.

sheralim012:
@geoo89 I have removed page-level caching, could you check now

geoo89:
LGTM
@istride Any concerns from you end?

Please merge in #1197 (or develop once #1197 has been merged into develop) so we can see if the tests are still succeeding.

geoo89:
Sorry, could you merge in the latest develop again? The cypress DB has conflicts. Once that's done, I will merge.

iogt/settings/base.py
```diff
@@ -14,6 +14,7 @@
 
 import allauth
 from django.contrib import auth
+from django.core.exceptions import ImproperlyConfigured
```
istride, line 17: I don't think this is used anywhere.

iogt/settings/base.py
```diff
@@ -428,3 +430,53 @@
     'wagtailcore.locale': ['language_code'],
     'iogt_users.user': ['username'],
 }
+
+SESSION_ENGINE = 'django.contrib.sessions.backends.cached_db'
+
+CACHE_BACKEND = os.environ.get('CACHE_BACKEND')
+if CACHE_BACKEND == 'db':
```
istride, line 437: Redis is our preferred cache backend, but my concern is that we may not be able to set up a Redis cluster in the production environment before this PR is merged and released, so a temporary solution _may_ be necessary.

Using `DatabaseCache` as a temporary solution is not very appealing to me because a new cache table needs to be created for every deployment.

Using `LocMemCache` creates an in-memory cache for each process, which could be quite memory-intensive and lead to unpredictable results due to caches not being synchronised.

The `FileBasedCache` may be best as a temporary solution because it only requires a readable and writable directory on the local filesystem, which would be automatically created if it does not exist. I'm assuming that it is shared by all processes and that it's performance would be worse than in-memory, but better than database caching. Would you agree?

docker-compose.yml
```diff
@@ -23,6 +24,9 @@ services:
       WAGTAILTRANSFER_SOURCE_BASE_URL: https://example.com/wagtail-transfer/
       WAGTAILTRANSFER_SOURCE_SECRET_KEY: wagtailtransfer-source-secret-key
       BASE_URL: 'http://localhost:8000'
+      CACHE_BACKEND: db
```
istride, line 27: We have a redis server available but we're using 'db' as the cache backend?

istride (changes requested):

iogt/settings/test.py
```diff
@@ -10,6 +10,7 @@
         'PASSWORD': os.environ.get('DB_PASSWORD'),
         'HOST': os.environ.get('DB_HOST'),
         'PORT': os.environ.get('DB_PORT'),
+        'CONN_MAX_AGE': int(os.environ.get('DB_CONN_MAX_AGE') or 0),
```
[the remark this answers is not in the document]
sheralim012: didn't put it in `base.py` as it uses SQLite3 by default

iogt/settings/base.py
```diff
@@ -428,3 +430,53 @@
...
+        },
+    }
+elif CACHE_BACKEND == 'redis':
+    CACHE_HOST = os.environ.get('REDIS_CACHE_HOST')
+    CACHE_PORT = os.environ.get('REDIS_CACHE_PORT')
+
+    CACHES = {
+        "default": {
+            "BACKEND": "django_redis.cache.RedisCache",
+            "LOCATION": f"redis://{CACHE_HOST}:{CACHE_PORT}/1",
```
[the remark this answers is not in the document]
sheralim012: done, 
also simplified this structure by using `CACHE_BACKEND`, `CACHE_LOCATION`, and `CACHE_TIMEOUT` environment variables and avoid using if else

iogt/settings/base.py
```diff
@@ -428,3 +430,53 @@
...
+            'TIMEOUT': 300,
```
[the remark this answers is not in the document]
sheralim012: done

iogt/settings/base.py
```diff
@@ -428,3 +430,53 @@
...
+            "OPTIONS": {
+                "CLIENT_CLASS": "django_redis.client.DefaultClient",
+                "IGNORE_EXCEPTIONS": True,
+            },
+        },
+        'renditions': {
```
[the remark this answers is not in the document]
sheralim012: it is suggested by wagtail to improve image lookup https://docs.wagtail.org/en/stable/advanced_topics/performance.html#caching-image-renditions

iogt/middleware.py
```diff
@@ -96,3 +99,35 @@ def process_response(self, request, response):
...
+        globals_.site_settings = SiteSettings.for_request(request)
+        globals_.theme_settings = ThemeSettings.for_site(site)
+        globals_.locale = locale
+        if not cache.get('svg_to_png_map'):
+            map = {}
+            for svg_to_png_map in SVGToPNGMap.objects.all():
+                map.update({
+                    (svg_to_png_map.svg_path, svg_to_png_map.fill_color, svg_to_png_map.stroke_color): svg_to_png_map,
+                })
+            cache.set('svg_to_png_map', map, timeout=300)
```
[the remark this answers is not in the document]
sheralim012: removed it as added configurable timeout env. var

iogt/settings/base.py
sheralim012: yes, you are right, removed it

docker-compose.yml
sheralim012: we can use whichever cache backend we want to use or test here

iogt/settings/base.py
sheralim012: in `FileBasesCache` each cache value will be stored as a separate value and the user running the server has to have read and write access for it to work, see [here](https://docs.djangoproject.com/en/4.0/topics/cache/#filesystem-caching)

istride: After discussion with @ChrisMarsh82, it has been decided to have two options for cache backend:

- Redis
- No cache (bypass)

If necessary, an intermediate option can be implemented later. We should be prepared to release the changes in this PR without Redis infrastructure in place.

sheralim012: if we don't specify any cache django by default uses Local-Memory Cache, see [here](https://docs.djangoproject.com/en/4.0/topics/cache/#local-memory-caching-1)

istride: I would like to have the option to not cache at all, just in case there is any trouble.

iogt/settings/test.py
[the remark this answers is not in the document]
istride: Is there any harm in adding it to `base.py`? I'm assuming that it would be ignored if SQLite is used. My aim is to reduce duplication.

iogt/settings/base.py
sheralim012: let's discuss in a quick call

docker-compose.yml
sheralim012: if we have decided to go with Redis or no-cache then I believe no-cache would be the default option

iogt/settings/test.py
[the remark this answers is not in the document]
sheralim012: sure, added

istride (approved):

docker-compose.yml
sheralim012: used dummy cache as default option

PR #1195, opened 2022-03-10
Feature/reduce docker image

istride:
Attempt to reduce the size of the docker container image to be run in production, for the sake of security, by removing build dependencies and dev-only Python packages.


istride:
@ehmadzubair @sheralim012 would you be able to cast your eyes over this PR at your earliest convenience? this is not urgent.

sheralim012 (approved):

PR #1197, opened 2022-03-13
Adding tests to footers, including PageLink pages

### LINKED ISSUE #1209: Update protocol for wagtail-transfer fork in requirements.txt
Yesterday, github dropped support for unencrypted `git://`, see section **No more unauthenticated Git** in https://github.blog/2021-09-01-improving-git-protocol-security-github/

This causes installing dependencies and automated tests to fail:

```
  Running command git clone --filter=blob:none --quiet git://github.com/IDEMSInternational/wagtail-transfer.git /tmp/pip-install-1cq9spi8/wagtail-transfer_9873ce5e916741a597ad9e6317dabc64
  fatal: remote error:
    The unauthenticated git protocol on port 9418 is no longer supported.
  Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
  error: subprocess-exited-with-error
```

I think the following line in `requirements.txt` is the culprit:
```
git+git://github.com/IDEMSInternational/wagtail-transfer.git@enh/show-error-for-unimported-referenced-pages#egg=wagtail-transfer
```

edmoss345:
Some further tests for items within the footers. Have added some PageLink items to test different variations. Have also added tests to check the image behaviour when icons/ images are pulled from the destination pages into the footer context.

fixes #1209


geoo89:
I have only one concern with adding these additional images:
Do we have the licence to put these on the public repository? In particular the SVGs, because I believe UNICEF purchased them to be used on their sites, but I don't think they can be put here (as we would be re-licencing them under a BSD licence, which we can't do). Maybe it is safer to put dummy images instead. The bug SVG for example is fine because it's CC0 (https://www.svgrepo.com/svg/34477/bug-free). The Youth dummy placeholder should be fine to.
@cbunicef What about the other images that I believe were taken from the V1 site?

edmoss345:
Understood, will delete and just use CC0 content

geoo89:
Thanks, also for fixing #1209!

I believe the piggy bank which has already been there needs to be removed as well due to licence issues.

geoo89:
What happened to `.github/workflows/config.yml` in your latest commit?

edmoss345:
> What happened to `.github/workflows/config.yml` in your latest commit?

No idea how that happened, will revise the last commit

PR #1201, opened 2022-03-15
Bug/svg to png map unique constraint

### LINKED ISSUE #1091: Error when retrieving SVGToPNGMap
After migrating a site from V1 to V2 there were duplicate entries in the SVGToPNGMap table. This caused an internal server error and therefore the site to become unavailable. I'm not sure how the entries in the table become duplicated.

There should be no reason why retrieving an image should cause a site to become unavailable, so more robust exception handling is required around the SVG to PNG process.

[svgtopngmap_bug.txt](https://github.com/unicef/iogt/files/7978268/svgtopngmap_bug.txt)

sheralim012:
Closes #1091 

We have a unique constraint on SVGToPNGMap with nullable fields
and unique constraint on nullable fields doesn't work as expected (one null isn't equal to another null)
So uses empty string for fill and stroke colors

for existing sites we will have to remove entries with nulls
so I have added a management command (clear_svg_to_png_map) to clear the SVGToPNGMap table (these will be auto-filled)

after new deployment run `python manage.py clear_svg_to_png_map`


ehmadzubair (approved):

istride:
My suggestion here would be to make `fill_color` and `stroke_color` not nullable and set a default value of "black" or "#000" or whatever 'null' is interpreted as now. I think this would be much simpler.

Presumably, there will be unique constraint violations if a process attempts to create a duplicate entry. Do we have adequate error handling to deal with this situation?

sheralim012:
currently, null is interpreted as the color(s) SVG has for both fill and stroke
and we don't save those colors in the database as there could be more than one color for fills and strokes

although setting stroke color to "black/#000" does work but not for fill color as it's none in a lot of cases
we can handle unique constraint violations gracefully as there is only one place we are doing this

sheralim012:
@istride I have updated the PR and set empty string as default value for fill and stroke colors

there shouldn't be any error other than integrity constraint but for safety, we can broaden the except block to handle all exceptions

home/management/commands/clear_svg_to_png_map.py
```diff
@@ -0,0 +1,9 @@
+from django.core.management.base import BaseCommand
+
+from home.models import SVGToPNGMap
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        SVGToPNGMap.objects.all().delete()
```
istride, line 8: This script leaves the PNG files on the filesystem. If we're going to provide a handy script, I think it ought to be thorough. Please make sure that the changes to the database are committed before removing files from the filesystem.

home/models.py
```diff
@@ -1092,18 +1094,24 @@ class SVGToPNGMap(models.Model):
...
             except:
                 logger.warning(f"Failed to convert SVG to PNG, file={svg_path}")
                 return None
-            obj = cls.objects.create(
-                svg_path=svg_path, fill_color=fill_color, stroke_color=stroke_color, png_image_file=png_image)
+            try:
+                obj = cls.objects.create(
+                    svg_path=svg_path, fill_color=db_fill_color,  stroke_color=db_stroke_color,
+                    png_image_file=png_image)
+            except IntegrityError:
```
istride, line 1113: I think we ought to be super-paranoid here and catch any error, just in case. Please also log a warning message, so that we have some visibility on what's going on.

istride (changes requested):

home/management/commands/clear_svg_to_png_map.py
sheralim012: done

home/models.py
sheralim012: done, logging exception message too.

istride (approved):

PR #1202, opened 2022-03-15
simplify rapidpro user setup

### LINKED ISSUE #1099: Prod only: Chatbot CSRF issue
When RapidPro tries to send a message to our servers via the webhook, e.g. https://ec.iogt.site/en/messaging/api/rapidpro-webhook/, the authentication succeeds (if the credentials are wrong, we'd get a 401), however we get a 403 due to the CSRF token, see below.
This only happens in prod. It is not an issue in dev (see e.g. http://iogt-django-dev.idems.international/en/sections/internet-safety/do-you-know-the-person-adding-you/ ).

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta http-equiv="content-type" content="text/html; charset=utf-8">
  <meta name="robots" content="NONE,NOARCHIVE">
  <title>403 Forbidden</title>
  <style type="text/css">
    html * { padding:0; margin:0; }
    body * { padding:10px 20px; }
    body * * { padding:0; }
    body { font:small sans-serif; background:#eee; color:#000; }
    body>div { border-bottom:1px solid #ddd; }
    h1 { font-weight:normal; margin-bottom:.4em; }
    h1 span { font-size:60%; color:#666; font-weight:normal; }
    #info { background:#f6f6f6; }
    #info ul { margin: 0.5em 4em; }
    #info p, #summary p { padding-top:10px; }
    #summary { background: #ffc; }
    #explanation { background:#eee; border-bottom: 0px none; }
  </style>
</head>
<body>
<div id="summary">
  <h1>Forbidden <span>(403)</span></h1>
  <p>CSRF verification failed. Request aborted.</p>


  <p>You are seeing this message because this site requires a CSRF cookie when submitting forms. This cookie is required for security reasons, to ensure that your browser is not being hijacked by third parties.</p>
  <p>If you have configured your browser to disable cookies, please re-enable them, at least for this site, or for "same-origin" requests.</p>

</div>

<div id="explanation">
  <p><small>More information is available with DEBUG=True.</small></p>
</div>

</body>
</html>
```

### LINKED ISSUE #1109: RapidPro setup is tricky
The [current process for setting up the RapidPro integration][1] is a little bit cumbersome as it requires:

- setting username and password for the RapidPro chatbot user as environment variables
- running a script to create the chatbot user, using the env vars
- setting the user id of the chatbot user as an environment variable, so the app can identify the chatbot user

One suggestion would be to use user groups to designate the RapidPro chatbot user. The set up would then consist of running a setup script. The script itself would do the following:

- create group 'iogt_rapidpro_chatbot'
- create user (with auto-generated username & password)
- add user to group

Then, to identify the chatbot user, the app would get all users in the chatbot group, sorted by name, and take the top result (just in case there are multiple).

This would eliminate the need for env vars, making the setup process less prone to human error and possible for an app admin to do as there would be no need to set the designated chatbot user id as an env var (which can only be done by a system / infrastructure admin)

[1]: https://github.com/unicef/iogt/blob/develop/messaging/README.md#technical-set-up

sheralim012:
Closes #1109
Closes #1099 

- used token-based authentication instead of Basic auth which used username and password
- removed custom basic auth class and created custom permission for rapidpro group user
- removed rapidpro environment variables and simply hardcoded username
- created rapidpro group and user added to that can be used for communicating with rapidpro server
- any user belonging to rapidpro group can be used for communicating with repidpro servers
- tokens for all users belonging to rapidpro group will be displayed on chatbot channel listing page
- updated readme for rapidpro


istride (changes requested):
Just a few more changes.

- Remove setting `RAPIDPRO_BOT_USER_USERNAME`
- Default value for setting `RAPIDPRO_BOT_GROUP_NAME` should be 'rapidpro_chatbot', make it possible to override via env var.
- The setup script `sync_rapidpro_bot_user.py` should randomly generate a username (random UUID?) and, optionally, take a username via a command line flag.

iogt/settings/base.py
```diff
@@ -428,3 +426,13 @@
...
 }
+
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': (
+        'rest_framework_simplejwt.authentication.JWTAuthentication',
+    )
+}
+
+SIMPLE_JWT = {
+    'ACCESS_TOKEN_LIFETIME': timedelta(days=365),
```
istride, line 437: How did we decide what value to use here?

sheralim012: we can't set the token lifetime to infinity and don't have a mechanism to refresh the auth token from rapidpro so set it to 1 year and after 1 year we will have to update token on rapidpro server 
but we can override simplejwt package to bypass the expiration check for users belonging to `rapidpro_chatbot` group

sheralim012:
@istride changes addressed

istride (approved):

PR #1210, opened 2022-03-17
fixes size inconsistency for checkbox and radio for questionnaires

### LINKED ISSUE #1089: Radio buttons in Forms are different sizes in browsers on iOS
image

I've viewed this in Safari and Firefox, but I think this would be the same for all browsers on iOS as Apple requires rendering with Webkit

https://developer.apple.com/app-store/review/guidelines/
image

The text of the radio buttons is still clickable so functionality isn't broken.

sheralim012:
Closes #1089 

- fixes size inconsistency for checkboxes and radio for questionnaires
- this happens for safari browsers


geoo89:
@edmoss345

edmoss345:
Can't check explicitly as don't have access to a Mac, but believe fix is good as checkboxes sizing is now being set in CSS sheet, see below:

image

PR #1211, opened 2022-03-17
Fix help text for questionnaires

geoo89:
fixes #1191

image

image

PR #1213, opened 2022-03-17
sets helptext for date and datetime field on page load and field type change

### LINKED ISSUE #1147: Date entry on feature phones
On feature phones the date entry widget doesn't exist, so you have to enter your date in text format.
The accepted format isn't clear, however: 
image

Suggestion from #699:
> Tell users the format of data required in advance. For example, provide a sample of the format the data should take, such as DD/MM/YYYY.

I believe there was also an issue discussing dates via dropdown lists for day/month/year, but I cannot find it (is this how it is done in V1)?

sheralim012:
Closes #1147 

- on questionnaire edit automatically fill the helptext for date and datetime fields
- on fields type change to date and datetime automatically fill the helptext if empty
- if help text is non-empty then on questionnaire edit  or field type change then existing help text wouldn't be affected.
- update po files accordingly
- user can specify help text in any language and that will be honored
- on frontend translated string will be displayed

### Help Texts
- date: Date must be in this (YYYY-MM-DD) format
- datetime: Datetime must be in this YYYY-MM-DDTHH:SS format


ehmadzubair (approved):

geoo89:
Is this the user-facing help text? If yes, then it seems like we have two different mechanisms for this, probably we should be consistent here, see e.g. https://github.com/unicef/iogt/pull/1211/files

Also, the user-facing help text needs to be translated. After tagging the strings, you can run `manage.py translation_tracking` (see the instructions in README.md)

geoo89:
Re: my previous comment -- I think it's fine for now to have two different mechanisms (pre-fill vs leaving blank), can defer thinking about that to later.

However, the user-facing help text needs to be translated into the appropriate locale of the survey/quiz.

sheralim012:
@geoo89 could you check now

geoo89:
The language of the inserted text is the language the admin selected for the admin backend.
It should be language of the page that is being created/edited.

geoo89:
I also updated the translations sheets accordingly, could you merge this into your PR?
https://github.com/sheralim012/iogt/pull/6

geoo89:
While in the admin panel the text is shown in English, it will be translated when facing the user.

This is slightly confusing, but it's non-trivial to make it appear in the correct language also in the admin backend.

PR #1218, opened 2022-03-18
adds wolof lang support

### LINKED ISSUE #1214: Unknown language code wo
The Senegal site is not working due to internal server errors on pretty much every page (inc. admin). The pertinent error message is "KeyError: 'Unknown language code wo.'". An extract of the logs has been provided below.

The code 'wo' corresponds to [Wolof][1].

[unknown-language-code.txt](https://github.com/unicef/iogt/files/8287622/unknown-language-code.txt)

[1]: https://en.wikipedia.org/wiki/Wolof_language

sheralim012:
Closes #1214 

- adds support for wolof language


geoo89:
Could we change the display name to "Wolof (Not Yet Translated)"?

sheralim012:
@geoo89 yes, we can but it will show on the language switcher on the frontend (if locale detail is absent or active) and if and when we decide to remove it, we will have to do a deployment too.

geoo89:
Ok, then let's better leave it as is. Is there some place else we could make administrators aware?

sheralim012:
we can add a message in bold red color on the locale or locale detail page on the admin panel

ChrisMarsh82:
> we can add a message in bold red color on the locale or locale detail page on the admin panel

@sheralim012 Can we go with this solution.

sheralim012:
for Wolof
Screenshot 2022-03-24 at 12 40 14 PM

for all other
Screenshot 2022-03-24 at 12 40 27 PM

PR #1241, opened 2022-04-05
makes questionnaires fields required by default

sheralim012:


questionnaires/migrations/0028_auto_20220405_0956.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('questionnaires', '0027_auto_20220128_1047'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='surveyformfield',
+            name='required',
+            field=models.BooleanField(default=True, verbose_name='required'),
```
istride, line 16: @sheralim012 a few weeks ago we discovered that default values are enforced by the framework, not the database. i'm suspicious that this migration will effectively exactly re-create the table that already exists in the database. would you double-check whether this migration is actually necessary?

sheralim012: `AbstractFormField` had already a `required` field with default to `True`
but for `SurveyFormField` we changed its default to `False`

now we are simply altering the field default to `True`
so no new table will be created

PR #1242, opened 2022-04-05
adds missing migrations for manifest settings

sheralim012:


home/migrations/0045_auto_20220405_0956.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('home', '0044_auto_20220223_0819'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='manifestsettings',
+            name='language',
+            field=models.CharField(choices=[('ar', 'Arabic'), ('bn', 'Bengali'), ('ny', 'Chichewa'), ('en', 'English'), ('fr', 'French'), ('id', 'Indonesian'), ('kaa', 'Karakalpak'), ('km', 'Khmer'), ('rw', 'Kinyarwanda'), ('rn', 'Kirundi'), ('ku', 'Kurdish'), ('mg', 'Malagasy'), ('ne', 'Nepali'), ('nr', 'Ndebele'), ('pt', 'Portuguese'), ('qu', 'Quechua'), ('ru', 'Russian'), ('sn', 'Shona'), ('es', 'Spanish'), ('sw', 'Swahili'), ('tg', 'Tajik'), ('ti', 'Tigrinya'), ('ur', 'Urdu'), ('uz', 'Uzbek'), ('zu', 'Zulu'), ('xy', 'Testing')], default='en', help_text='Choose language', max_length=3, verbose_name='Language'),
```
istride, line 16: @geoo89 @sheralim012 does this mean that we need a database migration for each new language we add in the future?

sheralim012: with the current setup, yes
but we can use locale foreign key here to avoid this

PR #1243, opened 2022-04-05
Fix failing functional tests of page footers

istride:
Includes a modification to the SVGToPNGMap process to accommodate a non-standard
media root directory.


home/templates/home/tags/footer.html
```diff
@@ -7,16 +7,13 @@
         {% endfor %}
         {% if request.user.is_anonymous %}
             {% url 'account_login' as login_url %}
-            {% static 'icons/login.svg' as login_icon_path %}
```
geoo89, line 10: Is the reason to change this conciseness?

Does the static keyword have any side effects that we're missing out on my removing it? Just want to be sure this doesn't affect things in prod.

istride: The change is required because the cypress tests use `cypress_tests/media` as the `MEDIA_ROOT` directory and the SVG to PNG process uses the URLs of SVG files, which use the `MEDIA_URL` setting. The two settings can be different with `MEDIA_ROOT` being the root directory for media files on the local filesystem and `MEDIA_URL` being the base of public facing URLs to media files served by the app.

We don't want to serve SVGs directly but convert them to PNG first, so we need to find the SVG on the local filesystem, convert it then provide a URL to the PNG.

Having made the change for media files it is then necessary to treat static files in the same way, in case the `STATIC_ROOT` and `STATIC_URL` settings are not the same. The static tag gives the URL (`STATIC_URL`) of the file given to it, but what we need is the path to the file on the local file system (`STATIC_ROOT`). The SVG to PNG process only works at the moment because those two settings coincide with each other i.e. the root directory is called 'static' and the URL begins with 'static'.

sheralim012 (approved):

PR #1247, opened 2022-04-08
return assets urls for precache

### LINKED ISSUE #902: While browsing offline, CSS sometimes is not loaded
image

sheralim012:
Closes #902 

- added offline app page in sitemap
- added CSS files for precaching
- added js files for precaching
- added jsi18n for precaching
- added fonts files for precaching
- added image files for precaching
- added audio/video files for precaching
- served workbox files from server instead of googleapis
- removed unused empty manifest file


geoo89:
It appears that css, images etc do get cached now, as intended.
It seems like we're caching images in a higher quality than needed, if they have been previewed in the admin panel, for example, the sitemap contains:
```
    "/media/images/[...].original.png",
    "/media/images/[...].max-800x600.png",
```

It seems that none of the children of the Misc index page are getting cached (in particular that includes the pages with the offline button), so clicking the corresponding Footer gives an error. Similarly, the translation-not-found page isn't cached.

I haven't tested this on a phone, but in a private Chromium tab, I couldn't get it to recognize that the site is in offline mode. For example, footers to the login page were still visible, search wasn't greyed out, survey submit buttons were not greyed out. If I remember correctly, this was implemented, and looking at the PR you didn't change anything related to this, so maybe I have it set up incorrectly?

Finally, in the dev console, I got a few messages:
```
GET http://localhost:8000/es/jsi18n/ net::ERR_FAILED
GET http://localhost:8000/static/js/jquery-3.6.0.min.js net::ERR_FAILED
DevTools failed to load source map: Could not load content for http://localhost:8000/static/css/global/global.css.map: Connection error: net::ERR_CONNECTION_REFUSED
```
Not sure if they are relevant.

sheralim012:
- I remembered why I didn't filter out original images because the format we previously added to support the original size
Screenshot 2022-05-10 at 10 53 47 AM

- I have added offline app page in the sitemap too
- I have added jsi18n in sitemap too but I'm not getting jquery error, could you run `collectstatic` and then test again
- feature enabling and disabling while offline only works if the app is in PWA mode (it was done a while ago)

- You are right, translation-not-found isn't handled, I suggest we create a separate ticket for it as it isn't quite straightforward

geoo89 (approved):
All seems good to me now.

Happy for this to be merged, modulo the decision on the original size images (though that could be addressed in a future PR if we defer the decision).

PR #1254, opened 2022-04-13
Feat/clear user data mgmt cmd

### LINKED ISSUE #1252: Abilty to delete all user data
This option should only be available at the deployment level.

Need the ability to be able to take a backup from an existing database (i.e. Central) and use that to create a new deployment. We will then need the ability to remove all user data. Showing a single admin user as creating all the content.

sheralim012:
Closes #1252 

- Following data will be cleared by this management command
  - users and users' profiles
  - questionnaires submissions for polls, surveys, and quizzes
  - comments
  - comment flags (comment reporting)
  - chats
- this management command optionally accepts user ids that would be excluded from the clearing process (admins etc)
- ids for excluded users will be retained but their above-mentioned data will be cleared though

usage: 
- `python manage.py clear_user_data`: all users will be cleared
- `python manage.py clear_user_data --exclude-user-ids 1 2 3`: all users will be cleared except users with id 1, 2, and 3

### DIFF
--- a/home/management/commands/clear_user_data.py
+++ b/home/management/commands/clear_user_data.py
@@ -1,0 +1,18 @@
+from django.contrib.auth import get_user_model
+from django.core.management import BaseCommand
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--user-ids',
+            nargs="+",
+            required=False,
+            type=int,
+            help="IDs for users that wouldn't be cleared"
+        )
+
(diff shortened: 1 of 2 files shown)


geoo89:
Anonymous user submissions for surveys (i.e. where the user is None) are retained, I assume we want to get rid of those as well?

sheralim012:
> Anonymous user submissions for surveys (i.e. where the user is None) are retained, I assume we want to get rid of those as well?

all user submissions whether anonymous or not will be removed now

PR #1273, opened 2022-05-10
show clean name as disabled field for questionnaires

### LINKED ISSUE #1257: User-facing survey Pages retain some old settings even after they are changed
To reproduce:

1) Create a new survey
2) Create question 1, radio type, Label: "Test Survey 1 Q 1" Admin Label: "Test Survey 1 Q 1"
3) Publish survey
4) View the following in the user-facing HTML:

```
<div class="quest-item__content">
        <label class="quest-item__label quest-item__input--visible">
          <input type="radio" name="test_survey_1_q_1" value="1" id="id_test_survey_1_q_1_0">
            <span class="cust-check__title">1</span>
        </label>
    </div>
```

5) Edit question 1 to Label: "Test Survey 1 Q 1x" Admin Label: "Test Survey 1 Q 1x"
6) View the following in the user-facing HTML:

```
<div class="quest-item__content">
        <label class="quest-item__label quest-item__input--visible">
          <input type="radio" name="test_survey_1_q_1" value="1" id="id_test_survey_1_q_1_0">
            <span class="cust-check__title">1</span>
        </label>
    </div>
```

Although the visible label changes, the underlying names for the inputs do not change after the Label and Admin label are changed.

This impacts when eg making multiple copies of the some survey with slight variation.

Does this have any impact on the platform? If not then we can de-prioritize this bug.

sheralim012:
Closes #1257 

- show clean name (disabled) field in admin panel for questionnaires
- on save it is automatically filled by wagtail


geoo89:
It appears that this field is filled in on the first save (when it previously was blank), but not on subsequent saves, i.e. when it already has a value.

Thus, when changing the label, the clean name is not updated.

sheralim012:
@geoo89 this is expected behavior

ChrisMarsh82:
@sheralim012 Could we add a tag/label to the field saying "This will appear in the HTML but will not be visible to the user"

sheralim012:
@ChrisMarsh82 changes addressed

PR #1276, opened 2022-05-12
show warning of download data permission for group

### LINKED ISSUE #1234: Add text to clarify "Edit" permission
We need to clarify that the "Edit" permission for a Group allows to download the data from Forms, as well as edit Pages.

image

For the short term I'm fine adding a note above the Add a Page Permission button that "Giving a Group the EDIT permission will also allow them to download data for Polls, Surveys, and Quizzes for the Pages they can access."

Long-term we should have a separate button for permission to download data.

sheralim012:
Closes #1234 

- Show warning message for download data for group permission
Screenshot 2022-05-12 at 2 22 42 PM


edmoss345:
Have reviewed and message appears as expected, 

image

PR #1278, opened 2022-05-12
Enh/audit schema migrations

### LINKED ISSUE #1064: Audit schema migrations
To ensure the migrations don't import/load stuff that might change.

Example: `home/migrations/0026_auto_20211029_1634.py`
Is there a reason it imports `Survey` rather than using `apps.get_model`? If the Survey model changes in the future, the migration might fail.

sheralim012:
Closes #1064 

- Followed proposed solution # 2
- audit schema migrations
  - removes references of management commands
  - removes references of model access
- updates `create_initial_menu` management command to create flat menu items (previously they were being created in migration)
- `load_v1_db` mgmt cmd after data migration will call `create_initial_menu` mgmt cmd

new site setup
- run migrations: `python manage.py migrate` 
- create index pages
  - `python manage.py create_index_pages`
  - `python manage.py create_questionnaires_index_pages`
- create flat menu items: `python manage.py create_initial_menu`
- create rapidpro bot user: `python manage.py sync_rapidpro_bot_user`


istride:
Does this change how a new site is created? What should be the process now?

sheralim012:
@istride I have updated the description

istride:
Currently, creating a new site requires the following:

1) migrate
1) load_v1_db (optional)
1) load_v1_users (optional)
1) load_svg
1) load_logo_and_favicon
1) create_default_site
1) create_initial_menu
1) sync_rapidpro_bot_user (optional)

This PR will change this to:

1) migrate
1) load_v1_db (optional)
1) create_initial_menu (triggered by previous step)
1) load_v1_users (optional)
1) load_svg
1) load_logo_and_favicon
1) create_default_site
1) create_index_pages
1) create_questionnaires_index_pages
1) create_initial_menu
1) sync_rapidpro_bot_user (optional)

I have tested this new sequence without the v1 data migrations and it does not work. The problem occurs at 'create_initial_menu' (step 10) with the error message "Main language not found." @sheralim012 would you provide some options for fixing this?

I can't remember why 'create_initial_menu' needs to be triggered by 'load_v1_db', but it doesn't seem necessary now.

sheralim012:
> I have tested this new sequence without the v1 data migrations and it does not work. The problem occurs at 'create_initial_menu' (step 10) with the error message "Main language not found." @sheralim012 would you provide some options for fixing this?

> I can't remember why 'create_initial_menu' needs to be triggered by 'load_v1_db', but it doesn't seem necessary now.

sheralim012:
> I have tested this new sequence without the v1 data migrations and it does not work. The problem occurs at 'create_initial_menu' (step 10) with the error message "Main language not found." @sheralim012 would you provide some options for fixing this?

- the reason it doesn't work is that any previous step doesn't create the Locale Detail with main language (except content migration: it identifies main language from v1 and creates locale detail in v2)
- if main language isn't found then we can't create homepage (if it doesn't exist)
- we can simplify it by making a little assumption that homepages are already created and we just need to create menu items

it would be better if we jump on a quick call to discuss this
 
> I can't remember why 'create_initial_menu' needs to be triggered by 'load_v1_db', but it doesn't seem necessary now.

is was for automation, but I'll remove it

sheralim012:
@istride changes addressed

PR #1282, opened 2022-05-13
Enh/url arguments for ga

### LINKED ISSUE #1192: Add URL arguments to indicate key actions on the site which can be analyzed in Google Analytics
In V1, there were certain URL elements which were consistent with key actions on the site:

Successful form submission = "success" would be in the URL text
Search = "?q=" would be in the URL text
Comment submission = "comment_done" would be in the URL

We filtered for these texts in the URL as a simple way to track volume of activities in Google Analytics.

Can we add a function like this in V2 as a temporary stop-gap on the way to better GA integration?

Ideal would be

Successful form submission = "ga_formsubmit" argument in the URL
Search = "?q=" is fine, I think this should already be in place
Successful comment submission = "ga_commentsubmit" argument in the URL

Adding these arguments to the URL seems like the easiest solution to me. There is no need for the platform to interpret those arguments.

This should be solved in the longer-term through custom Google Analytics features like Events in GA4.

sheralim012:
Closes #1192 

-


istride:
@sheralim012 is this still relevant?

sheralim012:
> @sheralim012 is this still relevant?

yes, the reason it's still in the draft is because comment submission needs additional changes

istride:
@sheralim012 can this be closed, given the move to matomo?

sheralim012:
Not sure but I think this is irrespective of the analytics platform we use. If you think it's not relevant anymore, go ahead and close it.

sheralim012:
Not relevant anymore.

PR #1283, opened 2022-05-16
adds ga pixel tracking

sheralim012:


istride (changes requested):
I don't think we should remove the tag manager tracking, however, it should only show if the GTM tracking IDs are set.

Likewise, the GA tracking should only be present if the relevant site settings exist. I suggest that we could create our own tags for GTM and GA, and that they could be unit tested.

Please also address the failing tests.

iogt/settings/base.py
```diff
@@ -488,3 +489,7 @@
             'TIMEOUT': CACHE_TIMEOUT,
         },
     }
+
+GOOGLE_ANALYTICS = {
+    'google_analytics_id': os.getenv('GOOGLE_ANALYTICS_ID'),
```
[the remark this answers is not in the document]
sheralim012: you are right it isn't useful 
but the package expects that if no tracking code (initially site setting wouldn't have tracking code in our case) is given for the template tag then use this setting for it and if this setting is not available then throw an error

but with your suggestion that we write template tag ourselves than this issue wouldn't occur

sheralim012:
> I don't think we should remove the tag manager tracking, however, it should only show if the GTM tracking IDs are set.

ok, will add a check for it.

> Likewise, the GA tracking should only be present if the relevant site settings exist. I suggest that we could create our own tags for GTM and GA, and that they could be unit tested.

- ok, will add similar checks for it too

> Please also address the failing tests.

looking into it

docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
       WAGTAILTRANSFER_SOURCE_BASE_URL: https://example.com/wagtail-transfer/
       WAGTAILTRANSFER_SOURCE_SECRET_KEY: wagtailtransfer-source-secret-key
       BASE_URL: 'http://localhost:8000'
+      GOOGLE_ANALYTICS_ID: ''
```
istride, line 22: Unnecessary now.

istride (changes requested):
I would like the GTM tags to only show of the IDs are set in site settings.

Do you think it might be better to write template tags for this, and to unit test the tags?

docker-compose.yml
sheralim012: done

sheralim012:
> I would like the GTM tags to only show of the IDs are set in site settings.

made rendering of each setting conditional

> Do you think it might be better to write template tags for this, and to unit test the tags?

we can do that, but to test rendering/html we have been using cypress and moving forward we can use selenium
but if we want to write unit tests then I can do that too

istride (approved):

PR #1333, opened 2022-07-19
filter out already md5 generated files as static return those

sheralim012:
- django's `static()` function relies on static files manifest
- django's `static()` function accepts the normal file path and returns the hash attached URL
- previously I was passing md5 attached file in `static()` and it was throwing an error
- now only pass normal files to `static()` and filter out md5 attached files


istride (changes requested):
I think it's necessary to have automated testing of the regular expression. I would suggest creating a function/method that determines whether a given file is valid for inclusion as a 'static' file. This method can then be unit tested in isolation.

I also think it would be a good idea to [compile the regular expression](https://docs.python.org/3.8/library/re.html#re.compile) before using it.

PR #1335, opened 2022-07-20
Feat/push notification

sheralim012:
- Initial setup for push notifications
- user will receive a notification of a new message from the chatbot


home/processors.py
```diff
@@ -13,3 +13,9 @@ def show_footers(request):
     if start_link:
         show_footers_ = False
     return {'show_footers': show_footers_}
+
+
+def add_vapid_public_key(request):
```
istride, line 18: would it be better to keep this in the messaging module?

messaging/chat.py
```diff
@@ -35,6 +36,14 @@ def _record_message_in_database(self, sender, rapidpro_message_id, text, quick_r
...
                 message.text = f'{message.text}{text}'
                 message.save(update_fields=['text'])
+            else:
+                payload = {
+                    'head': 'New Message!',
+                    'body': 'Click here to view it.',
+                    'url': self.thread.get_absolute_url(),
+                }
+                for user in self.thread.users.all():
+                    send_user_notification(user=user, payload=payload, ttl=1000)
```
istride, line 46: Is it possible for `send_user_notification` to fail? Does it throw exceptions that we may need to handle?

istride (changes requested):
I'm worried by the amount of JavaScript that probably has no automated tests covering it. Is there any way to provide more coverage, for example, by adding Selenium tests?

Given that there is now much more JS in this project than before, we need to be more disciplined in our approach. What I see is a lot of global variables, which are likely to be problematic in future.

home/processors.py
sheralim012: done

messaging/chat.py
sheralim012: yes, added try-except block and log the exception

istride (approved):

PR #1337, opened 2022-07-21
Feat/per user questionnaire submission export

### LINKED ISSUE #860: The ability to export survey/poll/quiz data by user profile
Linked to #520

sheralim012:
Closes #860


questionnaires/tests.py
```diff
@@ -723,3 +728,577 @@ def test_questionnaires_detail_page_permission(self):
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data, [])
+
+
+class QuestionnairesTests(TestCase):
```
istride, line 733: The class name is not specific enough. I think we're testing features of the Admin UI here, so the name should reflect that.

There is also a lot of boilerplate and duplicated code, which makes it harder to understand what is happening in each test. Please tidy up this class to extract common code into utility functions/methods/classes if it makes sense.

questionnaires/views.py
```diff
@@ -20,3 +41,143 @@ def get_filename(self):
...
+    def get_filtering(self, for_form_pages=True):
+        """ Return filering as a dict for form pages or submissions queryset """
+        filter_name = 'usersubmission__submit_time' if for_form_pages else 'submit_time'
+        self.select_date_form = SelectDateForm(self.request.GET)
+        result = dict()
+        if self.select_date_form.is_valid():
+            date_from = self.select_date_form.cleaned_data.get('date_from')
+            date_to = self.select_date_form.cleaned_data.get('date_to')
+            if date_to:
+                # careful: date_to must be increased by 1 day
```
istride, line 86: if you agree, please change this comment to:
```
# date_to is advanced by 1 day to be inclusive of date_to in results
```

questionnaires/views.py
```diff
@@ -20,3 +41,143 @@ def get_filename(self):
...
+                result[f'{filter_name}__gte'] = date_from
+        return result
+
+    def render_to_response(self, context, **response_kwargs):
+        if self.is_export:
+            return self.as_spreadsheet(context['submissions'], self.request.GET.get('export'))
+        return super().render_to_response(context, **response_kwargs)
+
+    def get_context_data(self, **kwargs):
+        """ Return context for view """
```
istride, line 103: I think this comment is stating the obvious and is therefore not that useful or informative, so I suggest removing it. There are other comments like this one in this class and I suggest those are removed too, unless they would be valuable to readers in the future.

istride (changes requested):

questionnaires/views.py
sheralim012: used date filter instead of datetime, so don't need this comment now

questionnaires/views.py
sheralim012: ok

questionnaires/tests.py
sheralim012: done

istride (approved):

PR #1342, opened 2022-07-26
Bug/query params for ga

### LINKED ISSUE #1330: At least some URL parameters are not passed to Google Analytics
To replicate error in V2

1. Go to https://www.internetofgoodthings.org/en/?test
2. In [Google Analytics Realtime > Overview](https://analytics.google.com/analytics/web/#/realtime/rt-overview/a88936714w116819520p122182821/), note that the URL which appears is /en/ and not /en/?test

To see behavior in V1

1. Go to https://ci.goodinternet.org/?test
2. In [Google Analytics Realtime > Overview](https://analytics.google.com/analytics/web/#/realtime/rt-overview/a104378100w145476574p150206335/), note that the URL which appears is /?test=

`?test=` is also not ideal, as it should really be `?test`

I noticed that in the page source for V2 the GA call doesn't include the parameters:

URL I used:

`https://www.internetofgoodthings.org/en/?test`

From page source:

```
<img src="/en/djga/google-analytics/?&amp;p=%2Fen%2F&amp;tracking_code=UA-69353247-43" width="0" height="0" />
<img src="/en/djga/google-analytics/?&amp;p=%2Fen%2F&amp;tracking_code=UA-88936714-1" width="0" height="0" />
```

And from the djga repo, it looks like the path that is sent to Google Analytics is from that "p" parameter in the URL to the tracking pixel:

```
    # get the path from the referer header
    path = path or request.GET.get('p', '/')
```

https://github.com/praekeltfoundation/django-google-analytics/blob/ad5a0515575042058ed602af8d3f6d2def0ef608/google_analytics/utils.py#L75

```
    # build the parameter collection
    params = {
        'v': VERSION,
        'z': str(random.randint(0, 0x7fffffff)),
        'dh': domain,
        'sr': '',
        'dr': referer,
        'dp': quote(path.encode('utf-8')),
        'tid': account,
        'cid': visitor_id,
        'uip': custom_uip or client_ip,
        'ni': ni,
    }
```

https://github.com/praekeltfoundation/django-google-analytics/blob/ad5a0515575042058ed602af8d3f6d2def0ef608/google_analytics/utils.py#L94

Why doesn't this bug appear in V1? Maybe the middleware implementation somehow doesn't have the issue.

sheralim012:
Closes #1330 

- process and send query params to google analytics

### DIFF
--- a/home/templatetags/generic_components.py
+++ b/home/templatetags/generic_components.py
@@ -1,5 +1,4 @@
-from cgi import parse_qs
-from urllib.parse import urlparse, urlencode
+from urllib.parse import urlparse, urlencode, parse_qs
 
 from django import template
 from django.conf import settings
--- a/home/templatetags/generic_components.py
+++ b/home/templatetags/generic_components.py
@@ -1,5 +1,10 @@
+from cgi import parse_qs
+from urllib.parse import urlparse, urlencode
+
 from django import template
 from django.conf import settings
+from django.urls import reverse
+from google_analytics import CAMPAIGN_TRACKING_PARAMS
 
 import iogt.iogt_globals as globals_
 
@@ -84,3 +89,47 @@ def language_picker_style():
 @register.simple_tag
 def site_version():
     return settings.SITE_VERSION
+
+
+@register.simple_tag(takes_context=True)
+def google_analytics(context, tracking_code=None, debug=False):
+    if not tracking_code:
+        try:
+            assert settings.GOOGLE_ANALYTICS['google_analytics_id']
+        except KeyError:
+            return ''
+    # attempt get the request from the context
+    request = context.get('request', None)
+    if request is None:
+        raise RuntimeError("Request context required")
+    # intialise the parameters collection
+    params = {}
+    # collect the campaign tracking parameters from the request
+    for param in CAMPAIGN_TRACKING_PARAMS:
+        value = request.GET.get(param, None)
+        if value:
+            params[param] = value
+    # pass on the referer if present
+    referer = request.META.get('HTTP_REFERER', None)
+    if referer:
+        params['r'] = referer
+    # remove collected parameters from the path and pass it on
+    path = request.get_full_path()
+    parsed_url = urlparse(path)
+    query = parse_qs(parsed_url.query)
+    for param in params:
+        if param in query:
+            del query[param]
+    query = urlencode(query)
+    new_url = parsed_url._replace(query=query)
+    params['p'] = new_url.geturl()
+    params['tracking_code'] = tracking_code or settings.GOOGLE_ANALYTICS[
+        'google_analytics_id']
+    # append the debug parameter if requested
+    if debug:
+        params['utmdebug'] = 1
+    # build and return the url
+    url = reverse('google-analytics')
+    if params:
+        url += '?&' + urlencode(params)
+    return url
--- a/home/templatetags/generic_components.py
+++ b/home/templatetags/generic_components.py
@@ -1,4 +1,4 @@
-from urllib.parse import urlparse, urlencode, parse_qs
+from urllib.parse import urlencode, parse_qs, urlparse
 
 from django import template
 from django.conf import settings
@@ -115,7 +115,7 @@ def google_analytics(context, tracking_code=None, debug=False):
     # remove collected parameters from the path and pass it on
     path = request.get_full_path()
     parsed_url = urlparse(path)
-    query = parse_qs(parsed_url.query)
+    query = parse_qs(parsed_url.query, keep_blank_values=True)
     for param in params:
         if param in query:
             del query[param]
--- a/home/test.py
+++ b/home/test.py
@@ -1,12 +1,16 @@
+from urllib.parse import unquote, parse_qs
+
 from django.conf import settings
 from django.db.utils import IntegrityError
-from django.test import TestCase
+from django.template import Context
+from django.test import TestCase, RequestFactory
 from django.urls import reverse
 from rest_framework import status
 from wagtail.core.models import PageViewRestriction, Site
 from unittest.mock import patch
 
 from home.models import SVGToPNGMap
+from home.templatetags.generic_components import google_analytics
 from iogt_users.factories import UserFactory, GroupFactory
 from home.factories import ArticleFactory, HomePageFactory
 from wagtail_factories import SiteFactory
@@ -131,3 +135,24 @@ class SVGToPNGMapTests(TestCase):
                 fill_color='#555',
                 stroke_color='#666'
             )
+
+
+class GoogleAnalyticsTagsTestCase(TestCase):
+    def setUp(self):
+        self.request_factory = RequestFactory()
+
+    def test_query_param_without_value(self):
+        request = self.request_factory.get('/en/?test')
+        context = Context({'request': request})
+
+        rendered_template = google_analytics(context, tracking_code='my-code')
+
+        self.assertIn("/en/?test=['']", unquote(parse_qs(rendered_template)['p'][0]))
+
+    def test_query_param_with_value(self):
+        request = self.request_factory.get('/en/?test=abc')
+        context = Context({'request': request})
+
+        rendered_template = google_analytics(context, tracking_code='my-code')
+
+        self.assertIn("/en/?test=['abc']", unquote(parse_qs(rendered_template)['p'][0]))
--- a/iogt/templates/google_analytics_pixel_tracking.html
+++ b/iogt/templates/google_analytics_pixel_tracking.html
@@ -1,5 +1,4 @@
-{% load generic_components %}
-{% load wagtailsettings_tags %}
+{% load wagtailsettings_tags generic_components %}
 {% get_settings %}
 
 <div style="display:none">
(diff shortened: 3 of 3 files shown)


home/templatetags/generic_components.py
```diff
@@ -84,3 +88,47 @@ def language_picker_style():
 @register.simple_tag
 def site_version():
     return settings.SITE_VERSION
+
+
+@register.simple_tag(takes_context=True)
+def google_analytics(context, tracking_code=None, debug=False):
```
istride, line 94: I must assume that this is original code that hasn't been lifted from another project. In any case, this code should be fairly easy to write unit tests for and I would like to see some, please.

istride (changes requested):

home/templatetags/generic_components.py
sheralim012: added test cases

istride (commented):
Do we need `django_google_analytics_app` if we are copying and modifying the only part of it that we are actually using? Are we anticipating that we may use its celery-based features in the future?

If we are effectively taking ownership of the code for the google analytics tag then there must be more test coverage than this. I'll assume that we'll be adding more later when there's more time.

sheralim012:
> Do we need `django_google_analytics_app` if we are copying and modifying the only part of it that we are actually using? Are we anticipating that we may use its celery-based features in the future?

- yes, we still need that package 
  - we are only modifying the pixel tracking URL generation
  - that package is still doing processing and sending GA event
- yes, I believe we eventually want to move to middleware + celery solution

> If we are effectively taking ownership of the code for the google analytics tag then there must be more test coverage than this. I'll assume that we'll be adding more later when there's more time.

as mentioned above we aren't taking total ownership but just changing the params processing 
so we only need to add test cases for that and for the rest we can copy from the package, only if we decide to keep using the pixel tracking solution
if we choose to go with middleware + celery solution then we will have to added test cases for that

istride (approved):

cbunicef:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:w="urn:schemas-microsoft-com:office:word"
xmlns:m="http://schemas.microsoft.com/office/2004/12/omml"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Word.Document>
<meta name=Generator content="Microsoft Word 15">
<meta name=Originator content="Microsoft Word 15">
<link rel=File-List
href="file:///C:/Users/CHRISW~1/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<link rel=themeData
href="file:///C:/Users/CHRISW~1/AppData/Local/Temp/msohtmlclip1/01/clip_themedata.thmx">
<link rel=colorSchemeMapping
href="file:///C:/Users/CHRISW~1/AppData/Local/Temp/msohtmlclip1/01/clip_colorschememapping.xml">


<![endif]-->
</head>

<body lang=en-DK link="#0563C1" vlink="#954F72" style='tab-interval:36.0pt;
word-wrap:break-word'>


<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'>OK we need to adjust the format for how this is
working.<o:p></o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'>Here’s how I see the data in GA based on the URLs I
use:<o:p></o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>



URL | Visible in GA | Should be visible in GA | Other note
-- | -- | -- | --
https://sandbox.iogt.site/en/?test | /en/?test=[''] | /en/?test | =[‘’] should not be visible
https://sandbox.iogt.site/en/?test=some_text | /en/?test=['some_text'] | /en/?test=some_text | =['some_text'] should not be visible, should be   some_text
https://sandbox.iogt.site/en/   ?utm_source=source   &utm_medium=medium   &utm_campaign=name   &utm_id=id   &utm_term=term   &utm_content=content | /en/ | /en/ | Parameters do not appear in the URL (good) or in   campaign tracking (bad), which probably means the utm_ parameters are recognized   and stripped out by GA, but GA cannot process input like utm_source=[‘source’]
https://sandbox.iogt.site/en/   ?utm_source=source   &utm_medium=medium   &utm_campaign=name   &utm_id=id   &utm_term=term   &utm_content=content   &test | /en/?test=[‘’] | /en/?test | utm_ parameters are stripped out by GA and except &test   which is handled properly by GA
https://sandbox.iogt.site/en/?utm_source=source | /en/ | /en/ | GA strips out utm_ parameters even when they don’t   meet the requirements to be tracked
https://sandbox.iogt.site/en/search/?query=test | /en/search/?query=['test'] | /en/search/?query=test |  
https://sandbox.iogt.site/en/search/?query=test&query=test2 | /en/search/?query=['test',+'test2'] | /en/search/?query=test&query=test2 | IoGT shows the search results for “test2”, so I’m   guessing that IoGT uses the final parameter when there is a parameter defined   multiple times in a single URL   There appears to be no standard on whether to accept one or multiple repeated   arguments. uri   - Authoritative position of duplicate HTTP GET query keys - Stack Overflow   So in passing this to Google we should stay true to   the format of the original URL



<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'>Basically, right now we are reporting a URL with added
JSON back to GA, but we need to report a properly formatted URL with query
string. <o:p></o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'>Where there is no value under the key name, we should
return “keyname” rather than “keyname=”. Where there are multiple definitions
for a parameter we should repeat the key name, ie “query=test&amp;query=test2”.<o:p></o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'><o:p>&nbsp;</o:p></span></p>

<p class=MsoNormal><span lang=EN-US style='mso-ascii-font-family:Calibri;
mso-ascii-theme-font:minor-latin;mso-hansi-font-family:Calibri;mso-hansi-theme-font:
minor-latin;mso-bidi-font-family:"Times New Roman";mso-bidi-theme-font:minor-bidi;
mso-ansi-language:EN-US'>Parameters look like they are not re-sorted, eg alphabetically,
so that is fine.<o:p></o:p></span></p>


</body>

</html>

The pixel URL seems to match what I see in GA, so I think it's just a matter of adjusting how the pixel URL is generated.

ChrisMarsh82:
@sheralim012 queries without values are still represented as "?test=&test2=" rather than "?test&test2"

ChrisMarsh82:
@cbunicef It looks like the '=' is added automatically by Google Analytics. Even if it is sent without it is still added. We have tested this with https://ga-dev-tools.web.app/hit-builder/

PR #1346, opened 2022-07-27
serve offline content not found page for uncached pages

sheralim012:
- added an offline content not found page
  - it will be shown when a user tries to access a page that isn't cached
  - it will have a link to "Offline Content Index Page" 
- added offline content index page
  - it will be used to specify links to cached pages


iogt/views.py
```diff
@@ -109,6 +110,15 @@ def get(self, request, page_id):
         page = get_object_or_404(Page, id=page_id)
         pages = page.get_descendants(inclusive=True).live().specific()
         page_urls = []
+        active_locale = Locale.get_active()
```
istride, line 113: I am wondering why there are no corresponding changes to the tests.

iogt/templates/sw.js
```diff
@@ -16,7 +16,10 @@ self.addEventListener('activate', event => {
     event.waitUntil(self.clients.claim());
 });
 
-self.addEventListener('fetch', event => {
+
+const languageCodeRegEx = RegExp('^\\/(\\w+([@-]\\w+)?)(\\/|$)');
```
istride, line 20: i really wish there were tests for this regex.

home/models.py
```diff
@@ -494,6 +494,44 @@ class OfflineAppPage(AbstractArticle):
     ])
 
 
+class OfflineContentIndexPage(AbstractArticle):
```
istride, line 497: There's a lot of duplication here. `OfflineContentIndexPage` is very similar to `OfflineAppPage` which is very similar to `AbstractArticle`. Do we need a completely new model to deal with such minor differences? Is there no better way to express the differences than to copy 30+ lines of code each time?

istride (changes requested):

home/models.py
sheralim012: refactor to remove a lot of duplication

istride (changes requested):
Please rename the schema migration to how it was before.

In general, let's always be more descriptive with the file names of db schema migrations going forward.

sheralim012:
> Please rename the schema migration to how it was before.

done

> In general, let's always be more descriptive with the file names of db schema migrations going forward.

Django automatically generates the file names, but sure going forward will do that.


also I have reverted the removal of "Offline App Page", as it will remove existing pages (if any)
It would be better to handle it separately and inform country officers beforehand so they can manage existing pages

iogt/views.py
istride: what I mean here is that there are additional URLs that must be included, but there is no test to make sure that they are. Is there a good reason for not changing the tests?

sheralim012: I have added/updated tests cases to incorporate these cases too

istride (approved):

edmoss345:
Have tested on my local, creating some test sections/ articles for download and an associated 'Offline Content Index Page'. Have tested in a browser and in a web app. Taking the browser offline and killing the server and all works as expected, so overall lgtm

PR #1364, opened 2022-08-02
updates query params formatting and utm params parsing

sheralim012:
- fixes utm params parsing 
- fixes query params format


cbunicef:
This works perfectly, except that queries without values are still represented as "?test=&test2=" rather than "?test&test2"

PR #1365, opened 2022-08-03
show optional + given help text for question

### LINKED ISSUE #1191: In surveys (possibly polls/quizzes) the help text for non-required questions is "optional" rather than the intended help text eg "check all that apply"
"Optional" should show below the label and above the help text. It can be part of the help text block, with a newline between "optional" and the help text.

### LINKED ISSUE #1336: In surveys, help text is not appearing for single-line and multi-line text

sheralim012:
Closes #1336 
Closes #1191 
Screenshot 2022-08-04 at 10 14 49 AM

### DIFF
--- a/questionnaires/templates/poll/poll.html
+++ b/questionnaires/templates/poll/poll.html
@@ -24,13 +24,13 @@
                                 {% csrf_token %}
                                 {% for field in form %}
                                     <div class="quest-item" tabindex="0">
-                                        <div class="quest-item__header">
-                                            {% field_description field %}
-                                        </div>
                                         <div class="quest-item__desc">
                                             {% if field.widget_type != "checkbox" %}
                                                 <p>{{ field.label }}</p>
                                             {% endif %}
+                                        </div>
+                                        <div class="quest-item__header">
+                                            {% field_description field %}
                                         </div>
                                         {% render_field field %}
                                     </div>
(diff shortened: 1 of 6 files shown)


cbunicef:
@sheralim012 Since we are making this edit, I wonder if we could quickly resolve an outstanding issue from #1191 

Right now the components of the field are listed like this:

```
Optional (if applicable)
Help Text
Label
[Answer options]
```

The desired order is 

```
Label
Optional (if applicable)
Help Text
[Answer options]
```

The change would be to move Label to be above Optional and Help Text.

If this is a quick fix then it would be great to add it.

sheralim012:
I have addressed #1191 but there is an edge case of field type "Checkbox",

### Current design is 

Screenshot 2022-08-04 at 10 15 34 AM

Optional (if applicable)
Help Text
[Answer option] Label (side by side)

### New Design is

Screenshot 2022-08-04 at 10 45 29 AM

[Answer option] Label (side by side)
Optional (if applicable)
Help Text

@cbunicef which do you prefer, the current or new design for the checkbox?

my recommendation is to stick with the current design because the new design looks odd with multiple questions 

Screenshot 2022-08-04 at 11 28 28 AM

cbunicef:
I'm surprised Wagtail is so inconsistent for the chechbox format, but I understand why having the Label field in the database reflect the text next to the checkbox makes sense.

Can we keep the original format just for the checkbox type questions? I feel pretty strongly that the help text should be below the question for all other field types, especially it will be clearer for screen reader users.

PR #1376, opened 2022-08-17
MIgrate tests for footer from cypress to selenium

istride:


istride:
@sheralim012 @edmoss345 I've made an attempt at introducing 'Page objects' ([1], [2]) in the Selenium tests. Let me know if you think this would be useful, and any other comments you may have.

[1]: https://selenium-python.readthedocs.io/page-objects.html
[2]: https://www.selenium.dev/documentation/test_practices/encouraged/page_object_models/

edmoss345:
@istride looks good to me, use of the 'Page objects' seems like a good idea as should make everything more readable and repeatable. Understand how you've used them, but think I should be taking more advantage in the tests I am writing. Maybe once I've got a decent size PR of tests together (currently working on #1377 ), we can discuss some best practice for using the pages and modify my code as required

selenium_tests/homepage/test_footer.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertEqual(
+            section_item.title,
+            section.title,
+            'Section title is the link text'
+        )
+        section_page = section_item.click()
+        self.assertTrue(
+            section_page.url_matches(section.url),
+            'Clicking on footer link navigates to section page'
+        )
```
sheralim012, line 46: this looks good, but can you divide this into isolated test cases? like clicking on the button in a separate test case and checking for icon (if SVG is given then check it if not then check png if both are absent then nothing) etc.

selenium_tests/pages.py
```diff
@@ -0,0 +1,78 @@
...
+from selenium.webdriver.remote.webdriver import WebDriver
+
+
+class BasePage():
+
+    def __init__(self, driver: WebDriver) -> None:
+        self.driver = driver
+
+    def url_matches(self, path: str) -> bool:
+        return self.driver.current_url.endswith(path)
```
sheralim012, line 12: can we use [URL parsing](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse)?

sheralim012 (changes requested):
there are a few edge cases like the presence of SVG/PNG icons, page link page, and menu items, I believe those will be added later, right?

selenium_tests/homepage/test_footer.py
```diff
@@ -0,0 +1,78 @@
...
+    def setUp(self):
+        Site.objects.all().delete()
+        self.home = HomePageFactory()
+        self.site = SiteFactory(
+            site_name='IoGT',
+            hostname=self.host,
+            port=self.port,
+            is_default_site=True,
+            root_page=self.home
+        )
```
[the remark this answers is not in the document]
istride: I don't yet have a very clear idea of how to organise the creation of test data. In any case, I think it would be best to address this in a future PR - perhaps you can create one for your mixin idea?

selenium_tests/homepage/test_footer.py
istride: I agree that there is more that could be tested, but I think we don't get very much value from testing those details. I also think that those edge cases ought to be tested with lower-level tests. I could add those tests now, but as things stand I think we get the essence of the cypress tests without losing very much coverage.

selenium_tests/pages.py
istride: Good idea - this has been added.

sheralim012 (approved):

PR #1381, opened 2022-08-23
adds readme for notifications

### LINKED ISSUE #1482: Add/Update push notification configuration readme

sheralim012:
Closes #1482


istride (changes requested):
There's plenty of general information about VAPID in the first link, which is good, but it is a bit overwhelming for someone who just wants to set up the app for notifications.

Using keys generated on someone else's website is not acceptable, for security reasons. I would like to see a method that would be appropriate for production setups. For example, there are instructions for [generating keys with the openssl command](https://blog.mozilla.org/services/2016/08/23/sending-vapid-identified-webpush-notifications-via-mozillas-push-service/#ident), but I don't know how to apply the result of that as environment variables.

Instructions for developers using Docker Compose are necessary, but I would also like to see more general instructions for people (sys admins) not using Docker Compose and who might be deploying IoGT in a hosted environment.

I think it is desirable for VAPID keys to never be committed into git, so in order to avoid that I think developers using Docker Compose should add VAPID keys to their docker-compose.override.yml file.

There are three environment variables that need to be set:
```
VAPID_PUBLIC_KEY: ''
VAPID_PRIVATE_KEY: ''
VAPID_ADMIN_EMAIL: ''
```
What is each one for? What is the expected format?

sheralim012:
@istride could you check it now?

notifications/README.md
```diff
@@ -0,0 +1,12 @@
...
+- Notifications rely on [VAPID](https://web.dev/push-notifications-web-push-protocol/) keys
+- You can generate these keys using `openssl` as follows
+  ```
+  $ openssl ecparam -genkey -name prime256v1 -out private_key.pem
+  $ openssl ec -in private_key.pem -pubout -outform DER|tail -c 65|base64|tr -d '=' |tr '/+' '_-' >> public_key.txt
+  $ openssl ec -in private_key.pem -outform DER|tail -c +8|head -c 32|base64|tr -d '=' |tr '/+' '_-' >> private_key.txt
+  ```
+- Set `VAPID_PUBLIC_KEY` env. variable from `public_key.txt`
+- Set `VAPID_PRIVATE_KEY` env. variable from `private_key.txt`
+- Set `VAPID_ADMIN_EMAIL` env. variables accordingly
```
istride, line 12: What is the purpose of this setting? Who's email should be here? A Wagtail admin or a system administrator?

istride (changes requested):

notifications/README.md
sheralim012: System administrator, if push service provider find any problem they can inform us

[For reference: See here](https://blog.mozilla.org/services/2016/08/23/sending-vapid-identified-webpush-notifications-via-mozillas-push-service/#:~:text=subscription%20updates.-,Identifying%20Yourself,-Mozilla%20goes%20to)

istride: The pertinent part from that blog post is:

>...a mailto link for the administrative contact for this feed. It’s best if this email is not a personal email address, but rather a group email so that if a person leaves an organization, is unavailable for an extended period, or otherwise can’t respond, someone else on the list can. Mozilla will only use this if we notice a problem with your feed and need to contact you.

I assume that this does not only apply to Mozilla's push service but generally, no matter who provides service.

Would you update the readme to make it clear what the admin email is for?

sheralim012: done

istride (approved):

PR #1411, opened 2022-10-03
Bypass registration survey for admin users

### LINKED ISSUE #898: Users with Admin access should not be forced to complete registration survey before viewing Admin Panel
This is not a major issue but causes a confusing disconnect when users try to access Admin Panel and are redirected to public website and registration survey.

Alihassanc5:
Closes #898 

- Admin users were forced to take registration survey before accessing Admin Panel.
- Now Admin users can easily access Admin Panel without taking any survey.


iogt_users/tests.py
```diff
@@ -3,15 +3,19 @@
...
 class PostRegistrationRedirectTests(TestCase):
     def setUp(self):
         self.user = UserFactory(has_filled_registration_survey=False)
-        self.admin_user = UserFactory()
+        self.admin_user = AdminUserFactory(has_filled_registration_survey=False)
+        self.admin_access_permission = Permission.objects.get(codename='access_admin')
+        self.group = GroupFactory()
+        self.group.permissions.add(self.admin_access_permission)
 
         Site.objects.all().delete()
```
istride, line 20: I think we've already established that this should not be necessary in any test.

iogt_users/tests.py
```diff
@@ -3,15 +3,19 @@
...
 
 
 class PostRegistrationRedirectTests(TestCase):
     def setUp(self):
         self.user = UserFactory(has_filled_registration_survey=False)
-        self.admin_user = UserFactory()
+        self.admin_user = AdminUserFactory(has_filled_registration_survey=False)
+        self.admin_access_permission = Permission.objects.get(codename='access_admin')
+        self.group = GroupFactory()
+        self.group.permissions.add(self.admin_access_permission)
```
istride, line 18: I don't like that there are two different users being created in this common setup method but they are not both used in every test. I think this could be organised better:

- Set up the user in each test?
- Create a base class for common setup and a test class for regular users and another for admin users?

Only some variables need to be assigned to self e.g. `self.group` is only used in one test, `self.admin_access_permission` is not used outside of the setup method

istride (changes requested):
I think there is a distinction between different kinds of users that needs to be made clearer:

- superuser: has every permission
- admin: has `wagtailadmin.access_admin` permission
- regular: no admin permissions

We should consider creating factory classes for each of these, if they are likely to be used in other tests.

iogt_users/tests.py
```diff
@@ -3,19 +3,19 @@
...
 
-        Site.objects.all().delete()
-        self.site = SiteFactory(site_name='IoGT', port=8000, is_default_site=True)
-        self.home_page = HomePageFactory(parent=self.site.root_page)
+        root_page = PageFactory(parent=None)
+        self.home_page = HomePageFactory(parent=root_page)
+        self.site = SiteFactory(hostname='testserver', port=80, root_page=self.home_page)
 
         self.registration_survey = SurveyFactory(parent=self.home_page)
         self.site_settings = SiteSettingsFactory.create(
```
istride, line 21: `self.site_settings` is not used anywhere

iogt_users/tests.py
```diff
@@ -30,3 +30,22 @@ def test_user_locked_out_without_filling_registration_survey_form(self):
     def test_anonymous_user_can_browse_public_urls(self):
         response = self.client.get(self.home_page.url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_admin_without_filling_registration_survey_form(self):
+        admin_user = AdminUserFactory(has_filled_registration_survey=False)
```
istride, line 35: I think this should be renamed so that it refers to 'superuser' instead of just admin. Something like, `test_superuser_should_not_complete_survey`.

iogt_users/tests.py
```diff
@@ -30,3 +30,22 @@ def test_user_locked_out_without_filling_registration_survey_form(self):
...
+        response = self.client.get(self.home_page.url)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_user_with_admin_access_do_not_need_to_fill_registration_survey_form(self):
+        admin_access_permission = Permission.objects.get(codename='access_admin')
+        group = GroupFactory()
+        group.permissions.add(admin_access_permission)
+
+        self.user.groups.add(group)
```
istride, line 47: I thought this might be the kind of code that would be in an 'admin user factory', so that if you need an admin user there is a standard way to do it across the whole test suite. I'm happy to leave this as it is if you think it's a premature optimisation.

istride (changes requested):
I think there should be at least one more test - for a user who has completed the post-registration survey, who should not be redirected.

iogt_users/tests.py
Alihassanc5: Admin user already has 'wagtail admin access'. But in this test-case we are giving wagtail admin access to a simple user, so I think it should be as it is here.

iogt_users/tests.py
Alihassanc5: Fixed it!

iogt_users/tests.py
Alihassanc5: fixed it

iogt_users/tests.py
istride: "Superuser" has all permissions. "Admin user" has admin access permission. This is the distinction we're failing to make at the moment, but I think it might be useful. Let's leave it for now and see whether this comes up again in future.

istride (approved):

PR #1417, opened 2022-10-10
Questionnaires added to flat menu

### LINKED ISSUE #1363: Surveys (questionnaires?) added to Flat Menu via Explorer do not appear
To reproduce:

1. Edit a currently live Flat Menu
2. Add a new item, which is a link to a survey, using the Choose A Page function
3. Fill out other elements of the item
4. Save
5. Item does not appear in Flat Menu
6. Edit Flat Menu to "Clear" the selected Page, and replace with a relative URL to the survey, and save
7. Item does appear in Flat Menu

Alihassanc5:
Closes #1363 

- Previously questionnaires and pagelinkpage added to flat menu were not showing on web page.
- This problem is fixed by setting the default value of show_in_menu to True for all questionnaires and pagelinkpage. 

### Updating show in menu for existing pages
- Run "./manage.py update_show_in_menu"


sheralim012:
write management command to update show in menu for existing pages

home/management/commands/update_show_in_menu.py
```diff
@@ -0,0 +1,8 @@
+from django.core.management.base import BaseCommand
+
+from wagtail.core.models import Page
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        Page.objects.all().update(show_in_menus=True)
```
istride, line 8: I don't think its a good idea to update all `Page`s. I think it would be safer to limit updates to `Page`s where `show_in_menus` is true by default.

home/factories.py
```diff
@@ -121,16 +121,27 @@ class FooterIndexPageFactory(PageFactory):
...
     class Meta:
         model = BannerPage
 
+
+class FlatMenuFactory(DjangoModelFactory):
+    class Meta:
+        model = FlatMenu
+
+
+class IogtFlatMenuItemFactory(DjangoModelFactory):
```
istride, line 145: I wonder whether this is really helping us to create flat menu items. What would happen if we did not use a factory?

home/tests/tests.py
```diff
@@ -98,3 +102,35 @@ def test_media_block_translation_of_bengali_language(self):
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, f"উপরের ভিডিও দেখা না গেলে <a href=\"{self.en_article.body[0].value.url}\" download> এর পরিবর্তে এটা </a> ডাউনলোড করুন", count=1)
         self.assertContains(response, f"উপরের অডিও শুনতে না পেলে <a href=\"{self.en_article.body[1].value.url}\" download> এর পরিবর্তে এটা </a> ডাউনলোড করুন", count=1)
+
+
+class FlatMenuTest(TestCase):
```
istride, line 107: This may be one of those cases when a test is not necessary. I don't think this is providing much value to us, and the consequences of not catching a regression are not particularly difficult to deal with - the management command can be applied, or a manual workaround via the CMS UI. My opinion is that we should remove it.

README.md
```diff
@@ -166,6 +166,12 @@ This process updates PO files as necessary and compiles a list of strings that a
 **Remark:** The file `common/translation_utils/translations.csv` is NOT updated in the process.
 Even though it has a column "is in use", its data is currently not updated automatically, and similarly, new strings are not added automatically.
 
+
+## Updating show in menu for existing pages
```
istride, line 170: I want to keep the main readme fairly short and clear of information that won't be relevant to most people. Suggest we move this to a page in docs directory - perhaps a troubleshooting page - with a link to it from the main readme.

istride (changes requested):

home/tests/tests.py
Alihassanc5: Okay, we can go for a manual workaround via the CMS UI.

README.md
Alihassanc5: It's moved to a page in the docs directory with a link from the main readme.

home/management/commands/update_show_in_menu.py
Alihassanc5: Fixed it!

istride (approved):

PR #1418, opened 2022-10-11
Fix poll results for checkboxes

### LINKED ISSUE #1306: Poll results for checkboxes question type only show "Your answer" for one option
I've selected both 1 and 2 as answers, then submitted:

image

Alihassanc5:
Closes #1306 

- Previously poll results for check boxes question type only show "Your answer" for one option.
- Now it will show "Your answer"  for all selected boxes.

image
image

### DIFF
--- a/questionnaires/templatetags/questionnaires_tags.py
+++ b/questionnaires/templatetags/questionnaires_tags.py
@@ -141,9 +141,9 @@ def get_item(dictionary, key):
 
 
 @register.filter
-def get_value_from_querydict(querydict, key):
+def get_values_from_querydict(querydict, key):
     dictionary = dict(querydict)
-    return dictionary.get(key)[0]
+    return dictionary[key]
 
 
 @register.simple_tag
--- a/questionnaires/templates/poll/poll_landing.html
+++ b/questionnaires/templates/poll/poll_landing.html
@@ -21,11 +21,11 @@
                             {% for question, answers in results.items %}
                                 {% snake_case question as current_answer %}
                                 {% for answer, count in answers.items %}
-                                    <div class="cust-check cust-check--white polls-widget__item{% if answer == request.POST|get_value_from_querydict:current_answer %} cust-check--active{% endif %}">
+                                    <div class="cust-check cust-check--white polls-widget__item{% if answer in request.POST|get_values_from_querydict:current_answer %} cust-check--active{% endif %}">
                                         <div class="cust-check__title">
                                             <div class="cust-check__title-left">
                                                 <span>{{ answer }}</span>
-                                                {% if answer == request.POST|get_value_from_querydict:current_answer %}
+                                                    {% if answer in request.POST|get_values_from_querydict:current_answer %}
                                                     <span class="cust-check__subtitle">{% translate "Your answer" %}</span>{% endif %}
                                             </div>
                                             <div class="cust-check__title-right">


istride:
Please resolve the conflicts.

istride (changes requested):
This is an opportunity to move some logic out of templates and into whatever is building the context. The results could be presented to the template as a list of items, with (amongst other things) a boolean for each item indicating whether it is 'active' or 'checked'.

The templates might then be simple enough not to be part of the testing. Unit tests would assert the contents of the context rather than HTML, which would be simpler and easier to understand. Selenium tests could be used (if not already) to test presentation and/or interactivity.

questionnaires/models.py
```diff
@@ -622,7 +622,7 @@ def get_form_fields(self):
     def get_submission_class(self):
         return UserSubmission
 
-    def get_results(self):
+    def get_results(self, query_dict=None):
```
sheralim012, line 625: query_dict => data

questionnaires/models.py
```diff
@@ -665,11 +665,19 @@ def get_results(self):
                 for k, v in results[key].items():
                     results[key][k] = round(v/total_submissions, 4) * 100
 
-        return results
+        results_list = []
+
+        for question, answers in results.items():
+            current_answer = question.lower().replace(" ", "_").replace("__", "_").replace('?', '')
```
sheralim012, line 671: Double-check if we can use question.clean_name here instead

questionnaires/templates/poll/poll.html
```diff
@@ -41,23 +41,20 @@ <h1 class="title polls-widget__title">{{ page.title }}</h1>
...
-                                                <div class="cust-check__title-left">
-                                                    <span>{{ answer }}</span>
-                                                </div>
-                                                <div class="cust-check__title-right">
-                                                    <span class="cust-check__percent">{{ count|floatformat:"0" }}
-                                                        {% if result_as_percentage %}%{% endif %}</span>
-                                                </div>
-                                                <div class="cust-check__percent-item"
-                                                     style="width: {{ count |floatformat:"0" }}%"></div>
+                                {% for answer, count, selected in results %}
```
sheralim012, line 44: is_selected

questionnaires/templates/poll/poll.html
```diff
@@ -41,23 +41,20 @@ <h1 class="title polls-widget__title">{{ page.title }}</h1>
...
+                                    <div class="cust-check cust-check--white polls-widget__item">
+                                        <div class="cust-check__title">
+                                            <div class="cust-check__title-left">
+                                                <span>{{ answer }}</span>
                                             </div>
+                                            <div class="cust-check__title-right">
+                                                <span class="cust-check__percent">{{ count|floatformat:"0" }}
```
sheralim012, line 51: Can we round it in python as well?

questionnaires/templatetags/questionnaires_tags.py
```diff
@@ -140,12 +140,6 @@ def get_item(dictionary, key):
...
 
 
-@register.filter
-def get_values_from_querydict(querydict, key):
-    dictionary = dict(querydict)
-    return dictionary.get(key)
-
-
 @register.simple_tag
 def snake_case(text):
```
sheralim012, line 144: remove if not being used.

sheralim012 (changes requested):

questionnaires/models.py
Alihassanc5: fixed it!

questionnaires/templatetags/questionnaires_tags.py
Alihassanc5: fixed it!

questionnaires/templates/poll/poll.html
Alihassanc5: Yes, fixed it!

questionnaires/templates/poll/poll.html
Alihassanc5: fixed it!

questionnaires/tests.py
```diff
@@ -1023,3 +1023,30 @@ def test_csv_export_with_modified_questions(self):
...
+        self.assertEqual(len(results), 3)
+        self.assertEqual(results[0][0], 'c1')
+        self.assertEqual(results[1][0], 'c2')
+        self.assertEqual(results[2][0], 'c3')
+        self.assertEqual(results[0][1], 100.0)
+        self.assertEqual(results[1][1], 100.0)
+        self.assertEqual(results[2][1], 0.0)
+        self.assertTrue(results[0][2])
+        self.assertTrue(results[1][2])
+        self.assertFalse(results[2][2])
```
istride, line 1051: I find it awkward that `results` is a list of tuples rather than say, a list of dicts, which I think might be easier to work with. I won't ask for this to be changed now though.

istride (changes requested):

selenium_tests/polls_specific/test_poll_functionality.py
```diff
@@ -69,4 +69,30 @@ def test_result_percentage(self):
         results_page = QuestionnaireResultsPage(self.selenium)
         poll_result = results_page.get_poll_results_text()
         self.assertIn("67 %", poll_result)
-        self.assertIn("33 %", poll_result)
\ No newline at end of file
+        self.assertIn("33 %", poll_result)
+
+    def test_multi_select_option_results_for_checkboxes(self):
```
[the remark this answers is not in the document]
sheralim012: It's checking for the exact bug that was overlooked by everyone, and I believe it's worth keeping it because when we decide to refactor the poll processing, tests like these will save us from different kinds of ripples, and give us confidence in the refactor as well.

questionnaires/tests.py
sheralim012: Agreed, but I believe it's better to keep the PR relevant and to the point so it's not too difficult to review. Maybe we can add it to the backlog.

selenium_tests/polls_specific/test_poll_functionality.py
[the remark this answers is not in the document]
istride: @sheralim012  This test replicates `test_multi_select_option_results_for_checkboxes` (which covers the bug) and the other test in this class, which is why I'm not convinced that it's useful.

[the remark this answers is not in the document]
Alihassanc5: @istride The unit test is removed.

istride (approved):

Issue #1420, opened 2022-10-11
Dashboard Feature Review


cbunicef:
### Needed before for MVP approval

- [ ] We have "limit 10000" on our chart queries - does this cause problems when we have more than 1000 responses?
- [ ] Superset users that are NOT Superset admins but ARE Wagtail admins, must be able to create and access databases - @sheralim012 this is an issue I think I identified today during testing. Can non-admins create datasets? Do we need a helper account to create the datasets?

### Changes needed after MVP

- [ ] Must have by user permission structure when creating datasets
- [ ] Creating dashboard should automatically take the IoGT account email address as the Superset login
- [ ] Dashboard title via markdown block and total submissions like in this dashboard: https://superset-sandbox.iogt.site/superset/dashboard/5/
- [ ] Must incorporate answer options where there were no responses, via metadata
- [ ] Set "Bar Values" setting on Chart > Customize tab by default
- [ ] Allow for sorting bars by "choices" order in dataset

sheralim012:
> * [ ]  We have "limit 10000" on our chart queries - does this cause problems when we have more than 1000 responses?

the "ROW LIMIT" on chart is for data points upon which aggregation is applied not data rows

100k records are shown in the chart
<img width="2055" alt="Screenshot 2022-10-12 at 11 03 39 AM" src="https://user-images.githubusercontent.com/49383675/195263398-298e65ae-0d53-483b-b746-b39d98e3329f.png">

after limiting the rows on chart, data points for aggregation are reduced not data rows and still showing 100k records
<img width="2056" alt="Screenshot 2022-10-12 at 11 12 34 AM" src="https://user-images.githubusercontent.com/49383675/195264009-8a695d6d-7a91-495b-a7ca-e6552270ba47.png">

sheralim012:
> * [ ]  Superset users that are NOT Superset admins but ARE Wagtail admins, must be able to create and access datasets - @sheralim012 this is an issue I think I identified today during testing. Can non-admins create datasets? Do we need a helper account to create the datasets?

- there are multiple roles in superset see [here](https://superset.apache.org/docs/security/)
- admin and alpha users have access to generate dashboard, datasets, charts out of the box
- gamma users need additional access permissions see [here](https://superset.apache.org/docs/security/#managing-data-source-access-for-gamma-roles)
  - gamma users must be given access to a specific database with dataset and chart read and write permission
[...]

PR #1425, opened 2022-10-17
Enh/apache superset

sheralim012:
#1420 

- superset admin users will create dataset and give ownership to current user

### DevOps changes
- add the following env. variables

```
SUPERSET_USERNAME
SUPERSET_PASSWORD
```


istride (changes requested):
I think that the app should have a service account in Superset - an account that represents the app in Superset. This account would give the app permission to do whatever it needs to do to create dashboards. If this is what is intended with this PR then I think the settings should be called:

- SUPERSET_USERNAME
- SUPERSET_PASSWORD

I would also encourage access to dashboards (and datasets, etc) be given to groups (in Superset) rather than individual users. The groups can be managed within Superset to control the level of access.

On the Wagtail side, I'm not keen on requiring Wagtail admins to authenticate themselves with their Superset credentials to create dashboards in Superset. I think the ability to create dashboards from Wagtail should belong to a Wagtail group, to which, users will be added if they need to create dashboards.

istride:
>if we don't authenticate individual users then we can't create dashboards, charts, etc under those users.

I'm suggesting that we only create dashboards under the app's service account, but give access to other Superset users via groups. There would be no ownership change, so no need to ask for credentials.

istride (approved):
Feel free to merge if no other changes are required.

PR #1438, opened 2022-10-24
adds media in page tree api for offline access

### LINKED ISSUE #1379: Cannot play videos on the Offline app
While testing the download button and content on the app noticed that video content won't download like the rest of the content. Only once I managed to see one second of a video.

sheralim012:
Closes #1379 

- add media (audio/video) assets for offline access


home/models.py
```diff
@@ -94,6 +94,10 @@ def get_context(self, request):
     def get_image_urls(self):
         return self._get_stream_data_image_urls(self.home_featured_content.raw_data)
 
+    @property
+    def get_media_urls(self):
+        return self._get_stream_data_media_urls(self.home_featured_content.raw_data)
```
istride, line 99: I don't understand why this implemented in this way because `home_featured_content` doesn't have a 'media' block - this will return an empty list every time.

home/mixins.py
```diff
@@ -61,10 +62,25 @@ def _get_stream_data_image_urls(self, raw_data):
...
+    def _get_stream_data_media_urls(self, stream_data):
+        media_urls = []
+
+        for block in stream_data:
+            if block['type'] == 'media':
+                media = Media.objects.filter(id=block['value']).first()
+                if media:
+                    media_urls.append(media.url)
+
+        return media_urls
```
istride, line 74: This could be replaced with:

```
def collect_media_urls_from_streamfield(field):
    return [ block.value.url for block in field if block.block_type == 'media' ]
```

Where `field` is a StreamField.

home/mixins.py
```diff
@@ -61,10 +62,25 @@ def _get_stream_data_image_urls(self, raw_data):
...
+
     @property
     def get_image_urls(self):
         raise NotImplementedError
 
+    @property
+    def get_media_urls(self):
+        raise NotImplementedError
```
istride, line 82: Why raise an exception by default and not, say, an empty list?

Why treat images and media separately instead of having a method to collect all URLs?

iogt/views.py
```diff
@@ -121,10 +121,12 @@ def get(self, request, page_id):
                 page_urls.append(page.url)
         translation.activate(active_locale.language_code)
         image_urls = []
+        media_urls = []
         for page in pages:
             if isinstance(page, (HomePage, Section, Article, Poll, Survey, Quiz)):
```
istride, line 126: Why not just `PageUtilsMixin`?

questionnaires/models.py
```diff
@@ -280,6 +280,15 @@ def get_image_urls(self):
 
         return image_urls
 
+    @property
+    def get_media_urls(self):
+        media_urls = []
+
+        media_urls += self._get_stream_data_media_urls(self.description.raw_data)
```
istride, line 287: `description` wouldn't contain a 'media' block, so this seems unnecessary.

istride (changes requested):
Very confused about how this has been implemented.

sheralim012:
Addressed in #1612

Issue #897, opened 2021-12-13
Form "date" field doesn't restrict years to 4 digits


cbunicef:
![image](https://user-images.githubusercontent.com/30042226/145850209-519ea943-13d1-4943-a69c-2e2a00963faf.png)

Although it doesn't accept a long year as valid:

![image](https://user-images.githubusercontent.com/30042226/145850384-600fd8c5-7e72-428c-bba5-f9290c8e4c70.png)

PR #1443, opened 2022-10-27
Restrict date field years to 4 digits

Alihassanc5:


Alihassanc5:
Closes #897 

- Previously user can write year in date/date-time field up-to 6 digits.
- This problem is fixed by restricting the year in input date widget.

image


### Alternate Option
- We can also achieve this by changing the date widget
 
image

PR #1452, opened 2022-11-02
Enable longer/secure password on user account registration

Alihassanc5:


iogt/static/css/accounts.css
```diff
@@ -95,3 +95,8 @@ form input[type="checkbox"] {
     font-weight: 500;
     line-height: 14px;
 }
+
+.fieldWrapper input[type=checkbox] {
+    transform: translateY(3px);
```
istride, line 100: This seems weird to me. Why use transform instead of, say, margin or padding?

iogt_users/forms.py
```diff
@@ -32,10 +39,10 @@ class AccountSignupForm(SignupForm):
     def __init__(self, *args, **kwargs):
         super(AccountSignupForm, self).__init__(*args, **kwargs)
         self.fields.pop('email')
-        self.fields["password1"] = IogtPasswordField(label=_("4-digit PIN"), autocomplete="new-password")
+        self.fields["password1"] = IogtPasswordField(label=_("Enter a 4-digit PIN or a longer password"), autocomplete="new-password")
```
istride, line 42: I think this will have to be re-translated. I've asked whether this should be done now or as a separate issue.

istride (commented):
Any copy (labels, help text, etc) should be approved by @cbunicef first. Anything that needs to be translated should be identified and flagged as soon as possible.

sheralim012:
@cbunicef you can take the translations.csv from this to add translations

Issue #1453, opened 2022-11-02
Ability to customize resizing of images across platform


cbunicef:
As a PLATFORM ADMINISTRATOR I would like more control over how images are resized on my platform. For example, I might want all my images to show at a high resolution if I am publishing content to be read by users who have strong internet connections, or I am publishing content with small visual details that are important.

Suggested implementation - a setting in Site Settings where the default maximum width of images can be customized, which replaces the current default of 360px.

Alihassanc5:
@cbunicef there is a special case,
In Section Listing other than first content then the width of image is 180px. 

Possible Solutions:
- Divide the maximum width of image by 2. (e.g, first-content: 360px; other-content: 180px)
- Another setting to set the width of other images.

cbunicef:
The size of that image as it's shown to the user is tied to the CSS. If it's not too complicated, it would be better to have the CSS code decide the resized width of the image. Ie the CSS code says the image is 50% width, so the image is resized to 50% width of the standard width ie 360px -> 180px

However, to me that does sound complicated. The other good option is letting the user manually set that width and giving the help text

"On Section Pages each of the sub-Pages are listed with their title alongside their Lead Image. By default the Lead Image is 50% of the width of the page, so normally you can set this image size to half of the width for full images. For example if your full image width is 360px then this can be set to 180px. If you are using a custom CSS you might need to set this to another size."

PR #1471, opened 2022-11-25
Adds ability to customize resizing of images across platform

Alihassanc5:


Alihassanc5:
Closes #1453 

- Now we are able to minimize the code at one place.
- Admin is able to choose different size of images across platform.

Screenshot from 2022-11-25 15-34-43

home/migrations/0050_sitesettings_maximum_width.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('home', '0049_auto_20221018_0943'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='sitesettings',
+            name='maximum_width',
```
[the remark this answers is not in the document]
Alihassanc5: @cbunicef it's updated. Can you check please?

cbunicef (commented):
Thanks Ali, this update looks good

cbunicef:
@istride can we consider getting this into the build which upgrades to Wagtail 2.15?

home/templates/home/tags/image.html
```diff
@@ -0,0 +1,25 @@
+{% load static wagtailcore_tags wagtailimages_tags %}
+
+{% if width == 360 %}
```
istride, line 3: Is there no other way to do this than to have a condition for every possible width?

home/templatetags/home_tags.py
```diff
@@ -143,3 +143,17 @@ def render_redirect_from_with_help_text(field):
                       f'See "https://docs.wagtail.io/en/stable/editor_manual/managing_redirects.html" for more details.'
 
     return {'field': field, 'red_help_text': True}
+
+
+@register.inclusion_tag('home/tags/image.html', takes_context=True)
+def render_image(context, image, half_width=False, img_class=None):
+    width = SiteSettings.get_for_default_site().maximum_width
```
istride, line 150: Shouldn't this be `default_image_resize_rule_width`?

istride: This needs to be absolutely bullet-proof. If anything goes wrong (no site settings, no default site) we should fall back to a sensible default setting.

istride (changes requested):
There must be tests for this feature.

home/migrations/0050_sitesettings_maximum_width.py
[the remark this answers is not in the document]
istride: The Wagtail documentation is giving instruction on how to use the `image` tag, rather than advice on how to name fields. I agree that the field name should be more descriptive than `maximum_width`, but `image_maximum_width` is probably good enough.

istride:
@cbunicef The Wagtail upgrade is ready to be deployed, so it is too late. It is important to get this feature correct because it affects all images across the site. When this feature is ready, we will merge it and then schedule a release.

cbunicef:
Thanks @istride, this is the type of feedback I was looking for. Let's please be sure that this is in the next build after the upgrade.

home/templates/home/tags/image.html
sheralim012: Basically, wagtail's image tag can't handle variables and it needs a constant string for filter specs, for reference see [here](https://github.com/wagtail/wagtail/issues/2090).
Wagtail has made some progress around it but is not quite there yet to handle this completely.
For now, we are doing this through if...elif...else instead of monkey patching wagtail image tag
Furthermore, instead of doing this in multiple places, we wrote an abstraction over the wagtail's image tag to hide away this ugly if...elif...else structure.
Let me know your thoughts on this.

home/templatetags/home_tags.py
```diff
@@ -143,3 +143,18 @@ def render_redirect_from_with_help_text(field):
                       f'See "https://docs.wagtail.io/en/stable/editor_manual/managing_redirects.html" for more details.'
 
     return {'field': field, 'red_help_text': True}
+
+
+@register.inclusion_tag('home/tags/image.html', takes_context=True)
+def render_image(context, image, half_width=False, img_class=None):
+    site = SiteSettings.get_for_default_site()
+    width = site.image_maximum_width if site.image_maximum_width else 360
```
sheralim012, line 151: site.image_maximum_width if site else 360

home/tests/tests.py
```diff
@@ -55,3 +57,38 @@ def test_do_not_change_queryset_when_parent_page_is_not_a_section(self):
         pages_after = limit_page_chooser(pages_before, request)
 
         self.assertEqual(pages_after, pages_before)
+
+
+class ImageResizeTest(TestCase):
+    def setUp(self):
+        Site.objects.all().delete()
+        SiteSettings.objects.all().delete()
```
sheralim012, line 65: no need to delete site and site settings, we can use already existing objects

home/tests/tests.py
```diff
@@ -55,3 +57,38 @@ def test_do_not_change_queryset_when_parent_page_is_not_a_section(self):
...
+        self.site = SiteFactory(site_name='IoGT', port=8000, is_default_site=True)
+        self.site_settings = SiteSettingsFactory(site=self.site)
+        self.section = SectionFactory(parent=self.site_settings.site.root_page)
+        self.article = ArticleFactory(parent=self.section)
+
+    def test_default_image_maximum_width_360(self):
+        response = self.client.get(self.article.url)
+        soup = BeautifulSoup(response.content)
```
sheralim012, line 73: parsed_response

sheralim012 (changes requested):

istride:
I would recommend not working on this until a couple of issues are resolved:

- Image width options: I'm not sure whether these are guesses which might need to change as soon as the app is deployed - I would like us to be certain.
- Is this feature the best way to solve the problem or would use of `<picture>` be better? I don't think we should do both.

home/tests/tests.py
```diff
@@ -98,3 +98,35 @@ def test_media_block_translation_of_bengali_language(self):
...
+
+
+class ImageResizeTest(TestCase):
+    def setUp(self):
+        self.site = SiteFactory(site_name='IoGT', port=8000, is_default_site=True)
+        self.site_settings = SiteSettingsFactory(site=self.site)
+        self.section = SectionFactory(parent=self.site_settings.site.root_page)
+        self.article = ArticleFactory(parent=self.section)
+
+    def test_default_image_maximum_width_360(self):
```
sheralim012, line 110: test with at least one width size other than default

sheralim012 (changes requested):

istride:
@sheralim012 This has been merged without my approval. I would have appreciated the chance to verify that the changes I asked for have been made correctly. It was important to me that the values for the IMAGE_SIZE_PRESET setting be a name rather than a number, to make it very clear that there is a very limited range of valid values - clearer than a single comment in 'base.py'.

I'm not going to insist on more changes at this time.

PR #1475, opened 2022-11-30
All permissions are added in Groups

Alihassanc5:
image

### DIFF
--- a/iogt/templates/wagtailusers/groups/includes/formatted_permissions.html
+++ b/iogt/templates/wagtailusers/groups/includes/formatted_permissions.html
@@ -18,51 +18,51 @@
                 <th>{% trans "Change" %}</th>
                 <th>{% trans "Delete" %}</th>
                 {% if custom_perms_exist %}
-                <th>{% trans "Custom permissions" %}</th>
+                    <th>{% trans "Custom permissions" %}</th>
                 {% endif %}
             </tr>
         </thead>
         <tbody>
             {% for content_perms_dict in object_perms %}
-            {# We do not want non-admin users to affect users or groups - https://github.com/unicef/iogt/issues/1347 #}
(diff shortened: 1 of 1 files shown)


iogt/templates/wagtailusers/groups/includes/formatted_permissions.html
```diff
@@ -41,6 +47,21 @@ <h2>{% trans "Object permissions" %}</h2>
                                 {{ content_perms_dict.delete.checkbox.tag }}
                             {% endif %}
                         </td>
+                        {% if custom_perms_exist %}
+                        <td>
+                            {% if content_perms_dict.custom %}
```
istride, line 52: Could this be a redundant check? If `custom_perms_exist` is true, does that not imply `content_perms_dict.custom` exists? If not, should we really just be checking that `content_perms_dict.custom` exists?

istride (changes requested):

iogt/templates/wagtailusers/groups/includes/formatted_permissions.html
Alihassanc5: @istride it's wagtail code. Do we need to change this?

istride: It would have been helpful to explain a little why this is the preferred solution. As I understand it, we are overriding this template due to a previous issue (which still stands), and we are merely synchronising this template with the one in Wagtail 2.15 (which has changed).

I find the code is questionable, and I wouldn't want us to be complacent about it because it has been copied from upstream. In this case, everything considered, I think it's ok to leave it as it is.

I am now wondering about other templates we might have overridden and whether we have a process for identifying them and synchronising them with any upstream changes.

istride (approved):

PR #1476, opened 2022-11-30
Title extended across the column if there is no Lead Image

### LINKED ISSUE #1469: In Section Page, if a listed Page doesn't have a Lead Image then the Title text should be extended across the column

Alihassanc5:
Closes #1469 

image


iogt/static/css/iogt.css
```diff
@@ -2583,3 +2583,7 @@ input[type=checkbox], input[type=radio] {
     margin-bottom: 12px
   }
 }
+
+.text-align-center{
```
istride, line 2587: I suggest this be renamed to something that doesn't imply the stylistic choice that will be taken if there is no lead image. To keep our options open with regards to changing the look of IoGT (i.e. theming) we should just indicate that no lead image is available e.g. `.no-lead-image`.

Alihassanc5: Okay got it, it's fixed now!

istride (approved):

PR #1480, opened 2022-12-07
Adds submit button translations

### LINKED ISSUE #900: Submit button text defaults to English when creating eg survey in a non-English language

sheralim012:
Closes #900 

- Adds "Submit Button Text" translation
Screenshot 2022-12-07 at 3 49 45 PM


questionnaires/templates/questionnaires/tags/submit_button.html
```diff
@@ -3,7 +3,7 @@
 <button type="submit" class="cust-btn cust-btn--dark survey-page__btn questionnaire-submit-btn">
     <span>
         {% if fields_step.paginator.num_pages == fields_step.number %}
-            {{ page.submit_button_text }}
+            {% translate page.submit_button_text %}
         {% else %}
             {% translate "Next" %}
         {% endif %}
```
istride, line 9: I think this a good opportunity to remove this conditional from the template. My preference would be for the submit button label to be passed into the template via context. This would allow the logic to be tested by inspecting the context.

istride (changes requested):
Please resolve the conflict with the develop branch.

home/tests/tests.py
```diff
@@ -55,3 +57,35 @@ def test_do_not_change_queryset_when_parent_page_is_not_a_section(self):
         pages_after = limit_page_chooser(pages_before, request)
 
         self.assertEqual(pages_after, pages_before)
+
+
+class SurveyTranslationTest(TestCase):
```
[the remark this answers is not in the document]
sheralim012: My intention behind this was to check all fields for at least one language in the future (to keep this PR small and to the point). Doing that will ensure that these kinds of bugs don't go overlooked.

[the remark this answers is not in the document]
istride: I wonder whether this can be done more efficiently with a static code analysis tool, or perhaps there is a single test or script that could be run to test all fields are translatable. My preference would be to scrap this test.

questionnaires/templatetags/questionnaires_tags.py
```diff
@@ -68,7 +68,11 @@ def field_counter(form, forloop, form_length, fields_step, questionnaire):
 
 @register.inclusion_tag('questionnaires/tags/submit_button.html')
 def render_submit_button(fields_step, page):
-    return {"fields_step": fields_step, "page": page}
+    submit_button_text = page.submit_button_text
+    if hasattr(page, "multi_step") and page.multi_step and fields_step.paginator.num_pages != fields_step.number:
+        submit_button_text = "Next"
+
+    return {'submit_button_text': submit_button_text}
```
istride, line 75: This function is getting all of its information from the page, so why not move this to `Questionnaire`?

Is there a test for the logic in this function?

home/factories.py
```diff
@@ -121,6 +122,13 @@ class FooterIndexPageFactory(PageFactory):
     class Meta:
         model = FooterIndexPage
 
+
+class SurveyFactory(PageFactory):
+    title = factory.Sequence(lambda n: f'survey{n}')
+
+    class Meta:
+        model = Survey
```
istride, line 130: I think this should be in the questionnaire module. Dependencies should flow towards home, not the other way.

istride (changes requested):

home/factories.py
Alihassanc5: Removed this factory from home module.

questionnaires/templatetags/questionnaires_tags.py
Alihassanc5: This logic is used in surveys, quizzes, and embedded questionnaires. If we move it to the **questionnaire**, it will be repeated multiple times.

istride: I think we can work around that. What is required is a method to get the submit button text given a 'fields_step'. There can be a default implementation in QuestionnairePage (for Survey and Quiz), and an override in Poll.

There should be tests for this.

istride (changes requested):
@sheralim012 I appreciate the effort that's gone into this - I'm happy with the result, so I think it's been worth it. Just one more small addition to a test and this should be good to merge.

questionnaires/tests.py
```diff
@@ -1023,3 +1024,32 @@ def test_csv_export_with_modified_questions(self):
...
+        site = Site.objects.get(is_default_site=True)
+        self.en_home_page = site.root_page
+
+    def test_survey_submit_button_text_without_multi_step_enabled(self):
+        survey = SurveyFactory(parent=self.en_home_page, multi_step=False)
+        SurveyFormFieldFactory(page=survey, field_type='singleline')
+
+        self.assertEqual(survey.get_submit_button_text(), "Submit")
+
+    def test_survey_submit_button_text_with_multi_step_enabled(self):
```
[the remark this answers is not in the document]
Alihassanc5: Fixed it!

PR #1481, opened 2022-12-07
fixes media block text translation issue

### LINKED ISSUE #1462: Some Translation strings not working
At least one translation string is currently not working, possibly others:

> If you cannot view the above video, you can instead %(start_link)sdownload it%(end_link)s.

You can save the translation string, but there is no impact on the user facing website. This was tested on ec and uz.

I have tested some translation strings without variables, and some with variables, and so far the one above is the only one that seems to not impact the site when it's changed.

sheralim012:
Closes #1462 

- Media block text was being translated in python and we haven't hooked our database translations into it
- Moved media block rendering into Django templates where we have hooked our database translations
- Added updated translation strings 

Screenshot 2022-12-07 at 11 44 00 PM


home/templates/blocks/media.html
```diff
@@ -0,0 +1,23 @@
+{% load wagtailcore_tags i18n %}
+
+{% if value.type == 'video' %}
```
istride, line 3: My preference would be to avoid conditionals in templates, if possible, because it's likely to be harder to test. For example, I see three separate templates here:
- media.html
- media-video.html
- media-audio.html

The main media template would use a more specific template depending on the type, but the switching would happen based on naming convention. Otherwise, perhaps the template to use can be decided within `MediaBlock`, where it may be easier to test.

I don't expect this to be changed, I am just making my preferences known.

istride (changes requested):

home/tests/tests.py
```diff
@@ -55,3 +58,56 @@ def test_do_not_change_queryset_when_parent_page_is_not_a_section(self):
...
+            original="If you cannot listen to the above audio, you can instead %(start_link)sdownload it%(end_link)s.",
+            translation="উপরের অডিও শুনতে না পেলে %(start_link)s এর পরিবর্তে এটা %(end_link)s ডাউনলোড করুন",
+            language=bn_locale.language_code)
+
+        response = self.client.get(bn_article.url)
+        soup = BeautifulSoup(response.content)
+        video_text = soup.find("p", {"class": "article__content--video"}).text.strip()
+        video_link = soup.find("p", {"class": "article__content--video"}).find("a").get("href")
+        audio_text = soup.find("p", {"class": "article__content--audio"}).text.strip()
+        audio_link = soup.find("p", {"class": "article__content--audio"}).find("a").get("href")
```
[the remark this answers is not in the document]
sheralim012: There is another approach to writing integration/system tests for Django templates, it checks if correct values are available in the context (if necessary) and the response contains what is expected irrespective of the location (no need to drill down to exact HTML element). The issue with this approach is that it isn't as robust as the current approach. Both have their own pros and cons but let me know if we should move to this approach it will simplify our tests.
For reference:
- https://test-driven-django-development.readthedocs.io/en/latest/03-views.html
- https://stackoverflow.com/questions/10608681/how-to-test-render-to-template-functions-in-django-tdd
- https://github.com/wagtail/wagtail/blob/main/wagtail/users/tests/test_admin_views.py#L109

home/tests/tests.py
```diff
@@ -55,3 +58,56 @@ def test_do_not_change_queryset_when_parent_page_is_not_a_section(self):
...
+
+class MediaTranslationTest(TestCase):
+    def setUp(self):
+        self.site = Site.objects.get(is_default_site=True)
+        self.en_home_page = self.site.root_page
+        self.en_article = ArticleFactory(
+            parent=self.en_home_page,
+            body__0__media=MediaFactory(type='video'),
+            body__1__media=MediaFactory(type='audio'),
+        )
```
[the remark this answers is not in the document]
Alihassanc5: Fixed it!

home/templates/blocks/media.html
Alihassanc5: Fixed it!

home/tests/tests.py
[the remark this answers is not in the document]
istride: If it makes sense to use a different approach, then I'm happy for you to take it. I also don't think there is a one-size-fits-all solution. My problem was with how this approach had been implemented i.e. with lots of duplication.

home/templates/blocks/media.html
```diff
@@ -1,23 +1 @@
...
-        <audio preload="none" controls>
-            {% for source in value.sources %}
-                <source src="{{ source.src }}" type="{{ source.type }}">
-            {% endfor %}
-            {% translate "Your browser does not support audio playback." %}
-        </audio>
-    </div>
-    <p class='article__content--audio'>{% blocktranslate %}If you cannot listen to the above audio, you can instead {{ start_link }}download it{{ end_link }}.{% endblocktranslate %}</p>
-{% endif %}
+{% extends template %}
```
istride, line 1: Is this file still required, or can it be removed completely?

sheralim012: It's required and renders whichever template is specified in context.

istride (approved):

PR #1483, opened 2022-12-08
adds apache superset setup and configuration guide

sheralim012:


questionnaires/superset/README.md
```diff
@@ -0,0 +1,100 @@
...
+    - Click "Settings" menu on upper right corner
+    - Click on "List Users" menu item
+    - Click on "+" icon to add a new user
+    - Fill in 
+      - First Name: <some-name>
+      - Last Name: <some-name>
+      - User Name: <some-username>
+      - Is Active?: checked
+      - Email: <some-email>
+      - Role: Choose "Alpha" Role
```
istride, line 75: Giving the service account user the 'alpha' role is too permissive. From the [security section](https://superset.apache.org/docs/security/#alpha) in the superset docs:

>Alpha users have access to all data sources

We wouldn't want this in a multi-tenant Superset setup. I think it should be sufficient for the service account to be granted:

- the 'gamma' role
- 'IoGT reader' role: read-only access to the IoGT database
- a role that allows creating and editing datasets

Then, to view dashboards, a user would require:

- the 'gamma' role
- the 'IoGT reader' role

Would this work?

sheralim012: ok, 
I have tested with the gamma role for the service account and everything is working as expected.

I have updated the readme as well.

PR #1484, opened 2022-12-09
document release process

sheralim012:


istride (changes requested):
This is a good start.

I think it would be better if the release process is documented in a separate file in the `docs` directory, with a reference in the main README.

I would like to see more information about:

- Quality assurance: how do we know that a release is ready?
- Changelog / release notes: how do we know/communicate what is in a particular release?
- Versioning: we are currently using Semantic Versioning - are you suggesting a different method? How do we mark / tag a particular release?

Some other things to bear in mind:

- There isn't just one prod server - UNICEF alone has approx. 40 deployments in production
- IoGT may be used by other organisations or individuals than UNICEF
- The IoGT version is shown in the Wagtail admin; what's the best way to keep this version label updated?

sheralim012:
> * Quality assurance: how do we know that a release is ready?

I believe the release should be considered ready when no checks are failing on GitHub and the manual QA person gives a go-ahead.

> * Changelog / release notes: how do we know/communicate what is in a particular release?

I believe we can use GitHub's release feature for this and all changelogs and release notes will be in the same and obvious place.

> * Versioning: we are currently using Semantic Versioning - are you suggesting a different method? How do we mark / tag a particular release?

If Semantic Versioning is working great then we don't have to change it, it was just a suggestion to use the current date for releases as I find it easy to remember and use in discussions and meetings.

> 
> Some other things to bear in mind:
> 
> * There isn't just one prod server - UNICEF alone has approx. 40 deployments in production

If we wouldn't be updating all the prod servers at the same time (assuming CI/CD isn't in place) for various reasons then it would be better to deploy releases/tags. That way deployments on different servers will be more independent.

> * IoGT may be used by other organisations or individuals than UNICEF

I believe the above-mentioned method can handle this scenario as well.

> * The IoGT version is shown in the Wagtail admin; what's the best way to keep this version label updated?

If we choose the route of deploying releases/tags then we have to find a way to pass the release/tag to generate env from release or something like that.

istride:
I'm happy for you to add your suggestions to the document. I have a few other considerations for you to bear in mind.

>ready when ... manual QA person gives a go-ahead

Are there cases where we wouldn't need manual QA?

>... we can use GitHub's release feature for this ...

Sounds good. How would we use it? How would it fit into the release process?

>... we have to find a way to pass the release/tag to generate env from release or something like that.

Indeed, and I would be interested to know what the possibilities are here. Also consider that we could go the other way, that is, keep the version number in Python code and use that as the authoritative source for git tagging and exposing the version number in the app. In terms of the release process, how and when would we update the version number?

Currently, our deployment artifact is a Docker container image. We could (in future) push this image to a public place and allow deployers of IoGT to download it from there. Nevertheless, it seems a good idea to allow others to build IoGT by themselves too, in which case being able to target a particular release would be useful.

sheralim012:
> Are there cases where we wouldn't need manual QA?

As of now can't think of any case when manual QA isn't required but if such cases arise we should add in the release process.

> > ... we can use GitHub's release feature for this ...
> 
> Sounds good. How would we use it? How would it fit into the release process?

I believe QA will prepare release notes and share them with whoever does the deployment and the DevOps person create a release with these notes in GitHub.
 
> > ... we have to find a way to pass the release/tag to generate env from release or something like that.
> 
> Indeed, and I would be interested to know what the possibilities are here. Also consider that we could go the other way, that is, keep the version number in Python code and use that as the authoritative source for git tagging and exposing the version number in the app. In terms of the release process, how and when would we update the version number?

For POC I tried following
- I created an annotated tag (e.g., v1.0 or 12-01-2023)
- I checkout that tag as branch
- I exported the current branch name as env. variable
- I was able to use it as the site version

If we go with semantic versioning then we can follow [this](https://semver.org/), and define some project specific guidelines e.g., if database schema is changed then update minor version, if no database schema is changed then update patch version, if new feature is included then update major version and etc.

sheralim012:
Not relevant anymore.

PR #1503, opened 2023-01-18
Testing selenium with multiple browsers

edmoss345:
Bit of an experiment to try and add testing on firefox as well as chrome. @istride maybe one we can discuss, some of the tests need to be adapted due to differences between Firefox and Chrome. Need to consider the best way to run the tests in both browsers (if that is something that is desirable)


edmoss345:
will close as changes here have been merged into #1505

PR #1505, opened 2023-01-18
Moving to selenium 4 and adding layout tests

edmoss345:
Was looking to add some test around relative location of elements and this is easily done in Selenium 4, so have upgraded references to Selenium 4. Have included a basic relative location test for now but scope to increase coverage, need to consider how best to sample different configurations, @istride can discuss in our catch up

### DIFF
--- a/docker-compose.selenium.yml
+++ b/docker-compose.selenium.yml
@@ -30,20 +30,42 @@ services:
     ports:
       - "9200:9200"
 
-  selenium-hub:
-    image: selenium/hub:3.141.59-20210929
-    container_name: selenium-hub
-    ports:
-      - "4444:4444"
-
   chrome:
-    image: selenium/node-chrome-debug:3.141.59-20210929
-    volumes:
-      - /dev/shm:/dev/shm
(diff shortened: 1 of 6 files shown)


edmoss345:
@istride investigated the multiple browser testing a bit further with selenium but couldn't find a way to automate tests across multiple browsers. Have set up firefox and edge nodes in the docker-compose, but they are currently not being used in the tests. There must be a way to loop through all the tests running in each browser but I couldn't get anything to work. Think the thread below is on the right lines but couldn't figure it out.
https://stackoverflow.com/questions/11869431/selenium-parallel-browser-testing-in-python

Also as we discussed, some tests had to be modified for different browsers and I couldn't find a workaround for this so have just left everything running in chrome. Think this is probably the best solution for now. This PR upgrades to selenium 4 and includes some basic tests around layout and varying the window size, so probably still worth merging. Let me know any comments, thanks

istride (changes requested):

istride (approved):

Issue #1400, opened 2022-09-09
Remove double header in single column view for devices with fast scrolling and larger screens


cbunicef:
As a USER of a device with a small screen or slow scrolling, it is useful to have a repeat of the header in the middle of the page.

As a USER of a device with a larger screen and faster scrolling, a repeated header in the middle of the page is unnecessary and breaks normal design conventions.

Without a reasonable way to measure screen size or scrolling speed, we can break this down into simple vs smart devices - ie, is JavaScript enabled?

This is a single feature that was broken out of the list of features in #1395

PR #1507, opened 2023-01-19
fixes html structure for menu

### LINKED ISSUE #1506: #menu anchor on Homepage is missing in 2.15.6
The default flat menu setup which is default for sites includes a link to the anchor, these seem to be universally not working now.

sheralim012:
Closes #1506 

- There was a requirement to hide the double header (search and language dropdown) for JS-enabled devices see #1400 
- Due to the structure of the HTML this change broke the "Menu" button as the Menu button was connected to the double header which is hidden for JS-enabled devices now
- Restructure the HTML and make double header independent of it.


cbunicef:
Thanks a lot @sheralim012 for the quick action

istride (approved):

PR #1511, opened 2023-01-23
Make the limit on the number of form input fields configurable

### LINKED ISSUE #1509: 400 error when adding a radio field to a long survey - possibly due to too many fields?
We are receiving a 400 error when adding another radio type field to this survey: https://rw.goodinternet.org/admin/pages/828/edit/

Adding a single-line type field works fine. Copying the survey to a new Page and then adding a radio field also gives a 400 error.

Alihassanc5:
Closes #1509


istride (changes requested):
Please make this configurable at runtime. Keep the default of 1000, but allow an override via env var.

istride (approved):

PR #1513, opened 2023-01-23
Bloomly

Alihassanc5:


istride (changes requested):
The changes seem to be ok, but we agreed to merge this into a separate 'bloomly' branch rather than 'develop'.

sheralim012:
Yes, you are right, we wouldn't be merging this into develop and this was only for review & feedback, now we can close this PR. I have pushed a bloomly branch on unicef see [here](https://github.com/unicef/iogt/tree/bloomly)

PR #1515, opened 2023-01-23
Decrease loading time of load_po_files management command

### LINKED ISSUE #1514: load_po_files management command is slow
We have created a management command (`load_po_files`) that loads translation strings from PO files in the project into an app database via `translation_manager.manager.Manager.load_data_from_po`. We have also patched the provided `Manager` so that existing translation strings are not overwritten in the database.

It is intended that `load_po_files` be run every time the app is upgraded, and I have found that this makes the upgrade process quite slow, especially, as in our case, when there may be 40+ deployments to upgrade at the same time. In addition, I suspect that `load_po_files` puts the database under quite a lot of stress, and when the database is shared between several deployments, this makes me nervous about upgrading deployments simultaneously, in case the database is swamped and becomes unresponsive.

I have very roughly and unscientifically timed the original manager and the patched manager - using the `time` command on my development machine with an sqlite database. From a fresh database there is little difference, but there is a significant drop in performance with the patched manager after the first invocation of `load_po_files`.

After a brief investigation, I found that the patched manager is issuing many requests to individually create `TranslationEntry`, which could be issued in bulk. I have made amendments to the patched manager to improve its performance, and the results can be seen below. The result is a manager that loads PO files faster than the original manager with no difference between invocations on a fresh database vs ongoing.

# Original Manager

## From fresh database
./manage.py load_po_files  11.09s user 1.97s system 28% cpu 45.314 total
./manage.py load_po_files  11.59s user 2.00s system 20% cpu 1:06.19 total
./manage.py load_po_files  11.45s user 2.08s system 20% cpu 1:05.85 total
./manage.py load_po_files  11.78s user 1.92s system 21% cpu 1:04.11 total

## After first invocation
./manage.py load_po_files  6.98s user 0.29s system 93% cpu 7.743 total
./manage.py load_po_files  6.50s user 0.28s system 93% cpu 7.245 total
./manage.py load_po_files  6.26s user 0.22s system 93% cpu 6.964 total
./manage.py load_po_files  6.38s user 0.21s system 93% cpu 7.069 total

# Patched Manager

## From fresh database
./manage.py load_po_files  15.45s user 1.98s system 34% cpu 50.859 total
./manage.py load_po_files  15.23s user 1.93s system 34% cpu 50.465 total
./manage.py load_po_files  17.72s user 2.31s system 29% cpu 1:07.30 total
./manage.py load_po_files  16.00s user 2.24s system 34% cpu 53.139 total

## After first invocation
./manage.py load_po_files  10.40s user 0.80s system 50% cpu 22.068 total
./manage.py load_po_files  10.28s user 0.83s system 38% cpu 28.928 total
./manage.py load_po_files  10.05s user 0.82s system 39% cpu 27.756 total
./manage.py load_po_files  10.21s user 0.84s system 39% cpu 28.124 total

# Faster Patched Manager

## From fresh database
./manage.py load_po_files  2.98s user 0.21s system 75% cpu 4.238 total
./manage.py load_po_files  3.14s user 0.23s system 76% cpu 4.418 total
./manage.py load_po_files  3.06s user 0.24s system 76% cpu 4.328 total
./manage.py load_po_files  3.15s user 0.22s system 76% cpu 4.407 total

## After first invocation
./manage.py load_po_files  3.03s user 0.23s system 76% cpu 4.269 total
./manage.py load_po_files  3.08s user 0.21s system 77% cpu 4.257 total
./manage.py load_po_files  3.23s user 0.20s system 78% cpu 4.377 total
./manage.py load_po_files  3.07s user 0.17s system 76% cpu 4.255 total

istride:
- Replaces individual requests to the database to create `TranslationEntry` records with a bulk request.
- Uses `tdict` (apparently unused) as a cache

Closes #1514


iogt/patch.py
```diff
@@ -164,24 +163,21 @@ def store_to_db(self, pofile, locale, store_translations=False):
         else:
             locale_dir_name = get_locale_parent_dirname(pofile)
 
-        t = TranslationEntry.objects.filter(original=m.msgid, language=language, domain=domain).first()
-        if t:
-            if t.translation:
-                translations_to_keep.append(m.msgid)
-                continue
+        if tdict.get((m.msgid, language, domain)):
+            translations_to_keep.append(m.msgid)
```
sheralim012, line 167: There is a case that goes like this 

1. po file is loaded that has an empty translation for a string and a DB record is created
2. po file has been updated with the translation
3. po file is loaded again and it should update the existing DB entry (only if the translation is still empty in DB)

e.g., 
locale/ar/LC_MESSAGES/django.po
msgid "Remember me"
msgstr ""

when po file is loaded DB record is created 

locale/ar/LC_MESSAGES/django.po
msgid "Remember me"
msgstr "تذكرنى"

load po file again, the same entry created in the previous step should be updated.

This refactoring is creating a new entry. I hope this makes sense but do let me know if you have any questions.

sheralim012 (changes requested):

iogt/patch.py
istride: If a new entry is created, will the old (empty) entry be removed? If so, the result would be the same as updating. It would be better not to update because that can't be done in bulk.

Maybe I should write tests to make sure we're getting the result we intend.

sheralim012: Both removing old (empty) entries and creating new entries or updating old entries would work. BTW Django does provide bulk update functionality see [here](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-update). Yes, tests would be great.

istride:
Created some tests for loading PO files to DB. Are there any cases I may have missed?

With the tests I felt more confident about refactoring the code, so I have chosen to extend the existing translation manager rather than patch it, and moved the code into the home "app" - where the management command is.

sheralim012 (approved):

PR #1518, opened 2023-01-25
Title extended across the column if there is no Lead Image

### LINKED ISSUE #1469: In Section Page, if a listed Page doesn't have a Lead Image then the Title text should be extended across the column

Alihassanc5:
Closes #1469 

image


sheralim012:
Handled in an other way.

PR #1519, opened 2023-01-26
Extended paragraph width 100% conditionally

### LINKED ISSUE #1469: In Section Page, if a listed Page doesn't have a Lead Image then the Title text should be extended across the column

hashim-rana-hub:
Closes #1469 

- Title will take 100% width when the lead image isn't available for section, article in section listing and home page featured content.

### DIFF
--- a/home/static/css/global/global.css
+++ b/home/static/css/global/global.css
@@ -806,7 +806,7 @@ body.rtl {
 
 .article-card .img-holder, .section-card .img-holder {
   border-radius: 8px;
-  flex:0 0 50%;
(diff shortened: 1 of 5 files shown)


hashim-rana-hub:
@sheralim012 could you please review.

sheralim012:
Section listing
Screenshot 2023-01-26 at 2 55 43 PM
Home page featured content
Screenshot 2023-01-26 at 2 55 24 PM
Section listing with progress
Screenshot 2023-01-26 at 3 09 26 PM

istride (approved):

PR #1529, opened 2023-02-09
Adds experimental per-user form data export

Alihassanc5:


Alihassanc5:
@cbunicef can you please take a look into it?

cbunicef:
In the Experimental version the column "Admin Label" needs to be renamed to "Field".

Other than that, this looks great. thank you.

Alihassanc5:
@cbunicef I have noticed the warning message on the green background was not visible, so we have these three possible options; can you please let me know which one we should adopt?

cbunicef:
Option 4 looks great. This will only show up when the flag for this feature is set right?

Alihassanc5:
@cbunicef it will show up until we replace the form data with the experimental form data permanently.

cbunicef:
So the process for the transition should be:

- From now until the features are ready, only platforms that have the feature flag set should see 1) the experimental button 2) the notice
- Once the features are ready and for three months, all platforms should see 1) the experimental button 2) the notice
- After three months, the original form button changes function to show the new data forms and the notice is removed

docker-compose.yml
```diff
@@ -30,6 +30,7 @@ services:
       SUPERSET_USERNAME: ''
       SUPERSET_PASSWORD: ''
       PUSH_NOTIFICATION: 'enable'
+      EXPERIMENTAL_FORM_DATA: 'enable'
```
istride, line 33: Please don't add env vars to docker compose for options that are not required. It fills the docker compose file with options that are not relevant to most developers and obscures options which are important.

questionnaires/templates/questionnaires/form_data.html
```diff
@@ -71,6 +71,15 @@ <h1>
...
         </form>
+          {% if experimental_form_data %}
+            <span class="warning">
+			    <span class="svg-wrapper">
+                    <img src='{% versioned_static "icons/warning.svg" %}' />
+			    </span>
+                In 3 months, the format of form data files in XLSX and CSV will be altered and updated with the
+                implementation of experimental form data.
+		    </span>
+          {% endif %}
```
istride, line 82: Please format this properly.

Why use `versioned_static` instead of just `static`?

questionnaires/tests.py
```diff
@@ -880,6 +880,25 @@ def test_export_page_ids_filter(self):
...
+        for stream in response.streaming_content:
+            byte_response += stream
+        expected_response = \
+            f'Page ID,Page Title,Submission ID,Submission Date,Field,Value\r\n' \
+            f'{self.poll.id},{self.poll.title},{self.user_submission_01.id},{self.user_submission_01.submit_time},User,{self.user_01.username}\r\n' \
+            f'{self.poll.id},{self.poll.title},{self.user_submission_01.id},{self.user_submission_01.submit_time},URL,{self.poll.full_url}\r\n' \
+            f'{self.poll.id},{self.poll.title},{self.user_submission_01.id},{self.user_submission_01.submit_time},{self.poll_question.clean_name},c1\r\n' \
+            f'{self.quiz.id},{self.quiz.title},{self.user_submission_03.id},{self.user_submission_03.submit_time},User,{self.user_01.username}\r\n' \
+            f'{self.quiz.id},{self.quiz.title},{self.user_submission_03.id},{self.user_submission_03.submit_time},URL,{self.quiz.full_url}\r\n' \
+            f'{self.quiz.id},{self.quiz.title},{self.user_submission_03.id},{self.user_submission_03.submit_time},{self.quiz_question.clean_name},c3\r\n'
```
istride, line 897: Can we wrap this in parentheses instead of using trailing backslashes at the end of every line?

questionnaires/wagtail_hooks.py
```diff
@@ -5,7 +5,8 @@
 from wagtail.core import hooks
 from wagtail.core.models import Locale
 
-from questionnaires.views import FormPagesListView, FormDataPerUserView, generate_dashboard
+from questionnaires.views import FormPagesListView, FormDataPerUserView, generate_dashboard, \
+    ExperimentalFormDataPerUserView
```
istride, line 9: Wrap in parentheses instead of using trailing backslash.

questionnaires/views.py
```diff
@@ -183,6 +184,23 @@ def get_filename(self):
...
+            'User': item.user.username,
+            'URL': item.page.full_url,
+        }
+        for clean_name, answer in json.loads(item.form_data).items():
+            data.update({
+                form_fields_dict.get(item.page_id, {}).get(clean_name, clean_name): answer,
+            })
+
+        return [dict(zip(self.list_export, [item.page_id, item.page.title, item.id, item.submit_time, field, value]))
+                for field, value in data.items()]
```
istride, line 201: It seems to me that that only difference is the list of column headers (`list_export`) and their corresponding values.

I think the difference would be clearer if we extract the part that gets the column values:
```python
def get_values(self, item, field, value):
    return [item.page_id, item.page.title, item.id, item.submit_time, field, value]
```

And use it in `get_rows` in the appropriate place:
```python
def get_rows(self, item, form_fields_dict):
    # ...
    return [dict(zip(self.list_export, self.get_values(item, field, value)]))
                for field, value in data.items()]
```

Then we can just override `list_export` and `get_values()` for the new format.

istride (changes requested):
Also needs to be caught up with latest changes in `develop`.

PR #1541, opened 2023-02-15
adds jquery feature flag

### LINKED ISSUE #1536: Add include jquery flag
Add a flag to disable jquery being downloaded as a resource on specific deployments. 

It is understood that notifications and offline access will not work if jquery is not included.

sheralim012:
Closes #1536 

- Adds feature flag to enable/disable jquery
- Disabling jquery will result in disabling "offline access" and "push notification"
- Restructure and Separate out admin and public site custom JS code


istride:
If we are using jQuery for the purposes of ensuring cross-browser compatibility, I'm not sure that's necessary anymore. In any case, I think it's worth exploring the possibility, especially while there is not much Javascript in the app. I have created a [draft pull request][1] that includes a rewrite of 'iogt.js' that does not require jQuery. I think it may be a good idea to combine that pull request with this one so that we can switch between JS that does and does not require jQuery - hopefully, without losing any functionality.

[1]: https://github.com/unicef/iogt/pull/1585

iogt/settings/base.py
```diff
@@ -534,3 +535,4 @@
 SUPERSET_PASSWORD = os.getenv('SUPERSET_PASSWORD')
 
 PUSH_NOTIFICATION = os.getenv('PUSH_NOTIFICATION', 'disable') == 'enable'
+JQUERY = os.getenv('JQUERY', 'disable') == 'enable'
```
istride, line 538: Disable jQuery by default?

istride (changes requested):
@sheralim012 please also see other comments in PR about rewriting existing JS to not use jQuery.

iogt/settings/base.py
Alihassanc5: @istride JQuery is enabled by default. This expression compares the value of the JQUERY environment variable with the string 'enable'. If the value is 'enable', then the expression will evaluate to True. Otherwise, it will evaluate to False.

iogt/templates/base.html
```diff
@@ -80,6 +80,11 @@
 </main>
 
 {# Global javascript #}
+<script src="{% url 'javascript-catalog' %}"></script>
+<script>
+    // for JS enabled devices hide double menu
+    document.querySelector('.footer-head').style.display = "none";
+</script>
```
istride, line 87: Let's avoid inline JS if possible. Please move this back into 'iogt.js' and 'iogt-no-jquery.js'.

istride (changes requested):
I thought we were going to keep the feature flag to toggle jQuery on/off but use it to switch between the different versions of 'iogt.js'. The default should be to use the jQuery version.

istride (approved):

PR #1543, opened 2023-02-15
Bug/questionnaires skip logic errors for admin panel

### LINKED ISSUE #1489: Cannot create survey with skip logic
Trying to create a survey with some skip logic. When I try and make the page I get the warning message around validation errors, but there is no prompt in the form below saying where the error is coming from.

I believe I have set everything up correctly for the skip logic, @sheralim012 am I missing something obvious?

image

### LINKED ISSUE #1538: In 2.15.6, skip logic requires "Multi-step" to be selected, but that is not part of the validation rules
We should raise a validation error on the multi-step check box if 
1. There is skip logic used
2. Multi-step is not checked

"Multi-step is required to properly display questionnaires that include Skip Logic."

sheralim012:
Closes #1489
Closes #1538

- StreamBlock & StructBlock validation errors were handled generically before
- But in newer versions, wagtail has introduced a specific mechanism to standardize the StreamBlock and StructBlock validation errors
- Used the new mechanism for StreamBlock and StructBlock validation errors
- Raise validation error when skip logic is added and multi-step isn't enabled.

Screenshot 2022-12-23 at 3 31 28 PM


questionnaires/forms.py
```diff
@@ -144,6 +151,9 @@ def clean(self):
                 if self.clean_errors:
                     form._errors = self.clean_errors
 
+        if should_be_multi_step:
+            self.add_error("multi_step", "Multi-step must be enabled with skip logic.")
```
istride, line 155: This is not the text that is specified in #1538.

istride: I have asked, in the issue, whether we ought to activate multi-step rather than warn the user.

istride (changes requested):
Existing Selenium tests have been using a method of hiding the Selenium driver and low-level element selection code behind 'Page Objects', to keep the tests uncluttered. See [pages.py](https://github.com/unicef/iogt/blob/develop/selenium_tests/pages.py) for details. The implementation is far from perfect and could benefit from further development and refinement. I think it would be useful to create a Page Object for the admin panel tests. Any questions, let me know.

Alihassanc5:
@istride issues are resolved, could you take a look?

selenium_tests/admin/test_skip_logic.py
```diff
@@ -0,0 +1,94 @@
...
+        SurveyFormFieldFactory(label='Q2', page=self.survey01, required=True, field_type='checkbox',
+                               skip_logic=skip_logic)
+        SurveyFormFieldFactory(label='Q3', page=self.survey01, required=True, field_type='checkbox',
+                               skip_logic=skip_logic)
+
+        self.visit_url(reverse('wagtailadmin_pages:edit', args=(self.survey01.id,)))
+
+        admin_page = WagtailAdminPage(self.selenium)
+        admin_page.select_skip_logic('survey_form_fields-0-skip_logic-0-value-skip_logic')
+        admin_page.select_skip_logic_question("survey_form_fields-0-skip_logic-0-value-question_1", "3")
```
istride, line 55: This does not help us to easily understand what is happening in this test. There is a better way to do this, but for now, I would like to see something like:

```
admin_page.select_skip_to_question(question=0, answer=0, skip_to=3)
```

I don't want to see any details of the HTML page like CSS selectors and element IDs in the test - these are low-level details that should be hidden at this level.

questionnaires/forms.py
```diff
@@ -169,31 +177,51 @@ def add_stream_field_error(self, position, field, message):
...
+
+        for i, form in question_data.items():
+            if form.is_valid():
+                data = form.cleaned_data
+                for logic in data['skip_logic']:
+                    if (logic.value['skip_logic'] in [SkipState.QUESTION, SkipState.END]
+                            and not cleaned_data['multi_step']):
+                        return True
+
+        return False
```
istride, line 223: I don't feel comfortable calling `clean()` and `is_valid()` again because I don't think it is intended for it to be called more than once, and it doesn't seem necessary to do so here. I don't mind looping through existing, clean data, but I did want to avoid re-cleaning the data. I think this should be modified to something like:

```
def should_be_multi_step(cleaned_data, question_data):
    ...
```

That is, the method should be a class method that we pass the cleaned data to. This should simplify the method quite a bit.

In addition, please make sure to clean things up a bit. For example:

```
for i, form in question_data.items():
    ...
````

The key/index `i` is not used so we should loop through values rather than items.

istride (changes requested):

Alihassanc5:
@istride All the issues are fixed, could you take a look?

istride (approved):

PR #1545, opened 2023-02-15
updates page size for apis

### LINKED ISSUE #1537: Enable data API pagination by default
The data API should be paginated by default at an appropriate number - let's say 15, same as the Admin Panel. There should be the optional URL argument 'page_size' where the user can specify a huge number to essentially turn off pagination (already in place).

This applies to the questionnaires list and submissions list endpoints, and to future endpoints which are lists.

@istride @sheralim012 This seems like it may be very important for the current response.

sheralim012:
Closes #1537


istride (approved):

istride:
@sheralim012 ready to merge, but looks like there is a conflict.

PR #1549, opened 2023-02-20
Rendition table optimization

### LINKED ISSUE #1526: Consider exposing rendition table to allow for manual optimization of specific images
As an ADMIN, I want to be able to use external tools to compress specific images that don't compress well through the default Wagtail settings. This is important because even if I upload a manually compressed image to Wagtail, the CMS will decompress and then re-compress using default settings when creating renditions.

As an ADMIN, I want very few users to be able to access this feature. Even users that are normally considered Admins should not be able to access this feature by default. It should be individually enabled for each user. Perhaps all Admins can be given the permission to enable that feature for themselves, a la "Developer Options" on Android, but they need instructions on how to do it.

Alihassanc5:
Closes #1526

### DIFF
--- a/home/wagtail_hooks.py
+++ b/home/wagtail_hooks.py
@@ -13,7 +13,7 @@ from wagtail.core import hooks
 from wagtail.core.models import Page
 from wagtail.core.models import PageViewRestriction
 from wagtail.admin import widgets as wagtailadmin_widgets
-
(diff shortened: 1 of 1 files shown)


sheralim012:
@istride as we aren't pursuing this route, ok with you if I close it?

istride:
Fine by me.

PR #1553, opened 2023-02-22
Randomization between contents

### LINKED ISSUE #1547: Randomization between two pieces of content
As an ADMIN I would like to be able to send users randomly to Content A or Content B. I might use this for:
- A/B testing of content, with success depending on a feedback survey or user quiz scores
- Breaking very long surveys down into smaller parts which are randomly distributed to respondents - all questions get answers, but nobody has to fill out the complete very long survey

Minimum requirements:
- Randomization is spread evenly among several possibilities
- Option for the randomized element to have a fixed text
  - For example, if the randomized element is a Page Button, the Page Button should be able to have a fixed text even if the random destinations have different title texts

What I don't believe matters right now:
- Whether the randomization happens as the page is built (random URL is inserted into HTML) or it happens on click (a set URL directs to a random destination)
- What specific function we add this to - I could see the randomization possibly happening in a Page Button, Page Link Page, possibly other places

Alihassanc5:
Closes #1547

- A new **Random Page Button** block is introduced, which will serve wagtail pages randomly.
- Admin can add multiple pages.
- The design of this block is same as **Page Button Block**.

### Admin Panel
image

### Frontend
image

### DIFF
--- a/home/blocks.py
+++ b/home/blocks.py
@@ -191,10 +191,10 @@ class RandomPageButtonBlock(blocks.StructBlock):
 
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context)
-        button_pages = value.get('pages')
-        button_page = random.choice(button_pages).value if button_pages else None
+        live_pages = [page.value for page in value.get('pages') if page.value.live]
+        button_page = random.choice(live_pages) if live_pages else None
 
-        if button_page and button_page.live:
+        if button_page:
             context.update({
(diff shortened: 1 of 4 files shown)


cbunicef:
@sheralim012 @istride I've added this to the Current Release project in In Progress list, this is something that we'll need to solve relatively soon as the survey we will use it for is coming up.

sheralim012:
Push the branch in IoGT repo.

PR #1558, opened 2023-02-24
Updates Community Comment Moderation views and permissions

### LINKED ISSUE #1521: Community Comment Moderation permission setting changed in 2.15.6
I noticed that in 2.15.6 our permissions for Comment Moderation looks different:

 image

Is the “Can moderate comments” setting now for the user-side comment moderation? And what does the “can view” setting mean here?

Is there a way we can make this more clear, eg separating the Admin Panel comment moderating from the Community Comment Moderation again? (I think this is how it was pre 2.15.6) It would also be helpful to explain "Can view" more clearly.

Alihassanc5:
Closes #1521 

- changed the label of **Can Moderate** to **Can moderate via admin panel** .
image

- Added a new group permission as **Community comment moderation**
image


istride:
I would prefer to leave the default permissions for comments intact and only add (separately) new permissions for community comment moderation, which is a customisation we have made. I want to avoid unnecessary changes to how a third-party software package is working. I also disagree that permission to moderate comments via the admin can be considered 'community' comment moderation.

My suggestion is that the permissions UI should show:

- Comment
  - Can moderate comments
  - Can view
- Community comment moderation
  - Can moderate comments
  
@cbunicef does this provide a clear enough demarcation?

cbunicef:
I think I remember the reason I wanted to differentiate ADMIN PANEL vs PUBLIC SITE, is that the Admin Panel interface for "Community comment moderation" has more functionality/data visibility than the public site interface.

So if someone has the "Community comment moderation > Can moderate comments" permission you're suggesting

and then they are given Admin Panel permission

That single permission change, which is not related to "Community comment moderation" gives them more access to the "Community comment moderation" model than they had before.

This is my recollection, which could be wrong.

cbunicef:
Per our discussion here is the wiki entry: https://github.com/unicef/iogt/wiki/Structuring-User-Groups-Permissions-to-Accommodate-User-Facing-Site-Permissions

comments/views.py
```diff
@@ -16,7 +16,7 @@
 from django.utils.translation import ugettext as _
 
 from comments.forms import AdminCommentForm, CommentFilterForm
-from comments.models import CannedResponse, CommentModeration
+from comments.models import CannedResponse, CommunityCommentModeration
 
 
 def update(request, comment_pk, action):
```
istride, line 22: This entrypoint needs to be better protected. I was able to anonymously approve and reject comments by crafting an appropriate URL.

istride (changes requested):

sheralim012:
Move "Can moderate comments on ADMIN PANEL" to "Comments" section

comments/wagtail_hooks.py
```diff
@@ -25,6 +27,19 @@ class XtdCommentAdmin(ModelAdmin):
...
+        Permission.objects.get_or_create(
+            codename='can_moderate_on_admin_panel',
+            content_type=content_type,
+            defaults={
+                'name': 'Can moderate comments on ADMIN PANEL',
+            }
+        )
+        permissions = super().get_permissions_for_registration()
+
+        return permissions.exclude(codename='can_moderate')
```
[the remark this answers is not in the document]
istride: In addition, there are other places where the 'can_moderate' permission is still used e.g. in templates.

istride (changes requested):
This needs quite a bit more work.

I don't see any tests and I think this work needs it, as we would want to make sure that there are no regressions. Manual testing hasn't served us very well, so far.

comments/migrations/0004_auto_20230512_0655.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.19 on 2023-05-12 06:55
```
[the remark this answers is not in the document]
Alihassanc5: We are using Django 3.2 at the moment.

istride:
Selenium tests found in `selenium_tests/articles/test_comments.py` have been recently skipped (in `develop`) because they are unreliable in Github Actions. We should review these tests in light of the changes in this PR.

istride:
Selenium tests found in `selenium_tests/articles/test_comments.py` have been recently skipped (in `develop`) because they are unreliable in Github Actions. We should review these tests in light of the changes in this PR.

comments/wagtail_hooks.py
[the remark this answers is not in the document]
istride: Would something like this work?

```python
p = Permission.objects.get(
    codename='can_moderate',
    content_type=ContentType.objects.get_for_model(self.model),
)
if p.name != self.CAN_MODERATE_LABEL:
    p.name = self.CAN_MODERATE_LABEL
    p.save()

return super().get_permissions_for_registration()
```

This might be preferable because then we won't have a 'hidden' permission, that existing users or groups may have, and that might take effect in unexpected ways when we upgrade the underlying comments app/library.

comments/models.py
```diff
@@ -112,33 +112,38 @@ def __str__(self):
...
+    state = models.CharField(max_length=255, choices=State.choices, default=State.UNMODERATED)
     comment = models.OneToOneField(
         to='django_comments_xtd.XtdComment', related_name='comment_moderation', on_delete=models.CASCADE)
 
     def __str__(self):
         return f'{self.comment.id} | {self.state}'
 
+    class Meta:
+        permissions = (
+            ("can_community_moderate", "Can moderate comments on PUBLIC SITE"),
```
istride, line 131: I know this is what @cbunicef suggested, but I'd like the all-caps ('PUBLIC SITE') to be changed to lower-case. I don't want to start a trend, and I feel it could be considered a bit patronising.

comments/filters.py
```diff
@@ -42,10 +42,10 @@ class ModerationFilter(SimpleListFilter):
...
     def lookups(self, request, model_admin):
         return [
-            (CommentModeration.CommentModerationState.UNMODERATED, 'Unmoderated'),
-            (CommentModeration.CommentModerationState.APPROVED, 'Approved'),
-            (CommentModeration.CommentModerationState.REJECTED, 'Rejected'),
-            (CommentModeration.CommentModerationState.UNSURE, 'Unsure'),
+            (CommunityCommentModeration.State.UNMODERATED, 'Unmoderated'),
+            (CommunityCommentModeration.State.APPROVED, 'Approved'),
+            (CommunityCommentModeration.State.REJECTED, 'Rejected'),
+            (CommunityCommentModeration.State.UNSURE, 'Unsure'),
```
istride, line 48: We have this in two separate places, and it feels like they will have to be synchronised with each other. This suggest that this should be defined in one place and referenced where it needs to be used.

Isn't the second item in these tuples just the same as the second item `State` itself e.g.:
```python
UNMODERATED[1] == 'Unmoderated'
```

comments/urls.py
```diff
@@ -1,17 +1,21 @@
...
-    path('community-moderation/', CommentsCommunityModerationView.as_view(), name='comments_community_moderation'),
+    path('comment/process-canned-response', views.ProcessCannedResponseView.as_view(), name='process_canned_response'),
+    path('community-moderation/', views.CommentsCommunityModerationView.as_view(), name='comments_community_moderation'),
+    path('comment/<int:comment_pk>/approve', views.ApproveCommentView.as_view(), name='approve_comment'),
+    path('comment/<int:comment_pk>/reject', views.RejectCommentView.as_view(), name='reject_comment'),
+    path('comment/<int:comment_pk>/unsure', views.UnSureCommentView.as_view(), name='unsure_comment'),
+    path('comment/<int:comment_pk>/hide', views.HideCommentView.as_view(), name='hide_comment'),
+    path('comment/<int:comment_pk>/publish', views.PublishCommentView.as_view(), name='publish_comment'),
+    path('comment/<int:comment_pk>/unpublish', views.UnPublishCommentView.as_view(), name='unpublish_comment'),
+    path('comment/<int:comment_pk>/clear-flags', views.ClearFlagsCommentView.as_view(), name='clear_flags'),
```
istride, line 20: I'd like to establish a convention for naming these paths. Should we prefix each one with the name of the app and/or model e.g. `comments_comment_approve` or `comment_approve`? Do we need to prefix at all? In any case, I'd like the naming to go from the general to the specific, rather than the other way around, as it is now.

You can leave the existing named paths as they are and focus just on the new ones we are introducing.

istride: I should say that I'm happy with the URL structure i.e. 'comment/\<id\>/\<action\>'.

istride (changes requested):
Much improved. The views need a bit more work, but I'd like that to happen once tests are in place that capture the new requirements.

Alihassanc5:
@istride test cases are included now, the process I have followed:

Tested single view with all users (Simple User, Community comment moderator and Admin comment moderator).
Conducted tests on all other views, ensuring that appropriate permissions were applied according to the user's role.

PR #1559, opened 2023-02-24
Update migration of Turkish Locale

Alihassanc5:


sheralim012:
Will be addressed via #1590

PR #1567, opened 2023-03-07
allow only specific columns in superset dataset

sheralim012:
Only allow specific columns in the apache superset dataset.

### DIFF
--- a/questionnaires/superset/datasets.py
+++ b/questionnaires/superset/datasets.py
@@ -16,7 +16,7 @@ class Dataset:
         }
 
     def put_body(self, columns, metrics):
-        sql = f"SELECT * " \
+        sql = f"SELECT id, form_data, submit_time, page_id, user_id " \
               f"FROM {self.table_name} " \
               f"WHERE page_id = {self.page_id}"
         return {
--- a/questionnaires/superset/utils.py
+++ b/questionnaires/superset/utils.py
@@ -43,6 +43,8 @@ CALCULATED_COLUMN_EXPRESSION_MAP = {
     'url': "trim(both '\"' from ((form_data::json)->'{}')::text)",
 }
 
+ALLOWED_COLUMNS = ['id', 'form_data', 'submit_time', 'page_id', 'user_id']
+
 
 class DashboardGenerator:
     def __init__(self, user, questionnaire, superset_username, superset_password):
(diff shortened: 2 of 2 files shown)


cbunicef:
Thanks Sher. I think it's also good that we are specifically listing the columns, that will prevent any new columns being included automatically.

questionnaires/superset/datasets.py
```diff
@@ -16,7 +16,7 @@ def post_body(self):
         }
 
     def put_body(self, columns, metrics):
-        sql = f"SELECT * " \
+        sql = f"SELECT id, form_data, submit_time, page_id, user_id " \
```
istride, line 19: Why not use `ALLOWED_COLUMNS` here? It would eliminate the need to change the column list in two separate places.

questionnaires/superset/utils.py
```diff
@@ -85,12 +87,14 @@ def _create_dataset(self, database_id, owner_id):
...
-            column.pop('type_generic', None)
-            column.pop('uuid', None)
+        columns = []
+        for column in copy(dataset_detail.get('result', {}).get('columns')):
+            if column.get('column_name') in ALLOWED_COLUMNS:
+                column.pop('changed_on', None)
+                column.pop('created_on', None)
+                column.pop('type_generic', None)
+                column.pop('uuid', None)
+                columns.append(column)
```
istride, line 97: This doesn't seem safe to me. It would be better to use a list comprehension with a function that extracts what we need from the response, creating a new column definition, e.g.

```
columns = [ for extract_column_definition(column) in dataset_detail.get('result', {}).get('columns') if column.get('column_name') in ALLOWED_COLUMNS ]
```

istride (changes requested):

questionnaires/superset/datasets.py
Alihassanc5: Fixed it.

questionnaires/superset/utils.py
Alihassanc5: Fixed it.

Alihassanc5:
@istride According to [this](https://github.com/apache/superset/pull/21840) a modification has been made to the API for creating virtual datasets and has been integrated, but the feature have not yet been released. We will update this accordingly after the release.

istride (approved):
@sheralim012 Eventually, it might be possible (and more Pythonic) to combine all the code  in 'questionnaires/superset' into a single superset.py file and that is the direction I prefer to go in. For now, please move `ALLOWED_COLUMNS` from 'constants.py' to 'questionnaires/superset/\_\_init\_\_.py' and update the imports accordingly.

Feel free to merge once done.

istride:
> ...a modification has been made to the API for creating virtual datasets ... but the feature have not yet been released. We will update this accordingly after the release.

@Alihassanc5 sounds good. I'll leave it up to you to keep track of that and create an issue when it's time to update.

PR #1574, opened 2023-03-14
Lead images featured when sharing on social media platforms

### LINKED ISSUE #1552: Lead Images don't get featured when sharing on WhatsApp, Telegram
Not sure if this is an issue across digital messengers, but when a Page is shared there is only a URL, or only a URL and a title. 

Some alternate examples, Reddit example is ideal for when we have a Lead Image, Gizmodo is ideal for when we do not:

image

This is the behavior pressing the Share button within Firefox on an iPhone.

Alihassanc5:
Closes #1552 

Social media platforms:

- Facebook / Facebook messenger 

IMG_20230314_162622

- Instagram

IMG_20230314_162639

- Whatsapp

IMG_20230314_162511

- Telegram

IMG_5F2DEA94072A-1

- Snapchat

IMG_20230314_162605

## Updated help text of SEO tittle tag and Meta Description:
image


Alihassanc5:
@cbunicef The feature is only added when a page has a lead image. Furthermore, in cases where there is no lead image, such as the example you quoted from Gizmodo, we can achieve this behavior by including a description such as 'we come from the future' in this case.

cbunicef:
Thanks for this quick work Ali. Do we have any option on the size of the image? I notice that in the screenshot in the PR the lead image as viewed on Telegram is small. However in the original issue, the lead image as viewed on the reddit link is large.

iogt/templates/base.html
```diff
@@ -27,6 +27,22 @@
...
+    <meta property="og:site_name" content="{{ page.get_site.site_name }}">
+    <meta property="og:title" content="{{ page.title }}">
+    <meta property="og:description" content="Internet of Good Things.">
+    <meta property="og:type" content="article">
+    <meta property="twitter:site" content="{{ page.get_site.site_name }}">
+    <meta property="twitter:title" content="{{ page.title }}">
+    <meta property="twitter:description" content="Internet of Good Things.">
+    <meta property="twitter:card" content="summary_large_image">
+    {% if page.lead_image %}
+        {% image page.lead_image original as lead_image %}
```
istride, line 40: The image size should be one of the sizes used elsewhere on the site.

iogt/templates/base.html
```diff
@@ -27,6 +27,22 @@
     <meta name="description" content=""/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
     <meta name="theme-color" content="#493174">
+    <meta property="og:url" content="{{ page.url }}">
+    <meta property="og:site_name" content="{{ page.get_site.site_name }}">
+    <meta property="og:title" content="{{ page.title }}">
+    <meta property="og:description" content="Internet of Good Things.">
```
istride, line 33: Description of the article/page or the site? Can we do better than 'Internet of Good Things'?

istride (changes requested):
Consider isolating these meta tags into a separate template/tag.

iogt/templates/base.html
Alihassanc5: @istride, there may be some articles or pages that do not have descriptions, it's not possible to create dynamic content for them. Do you have any suggestions for a term that could be used instead of 'Internet of Good Things'?

Alihassanc5:
@cbunicef help text of SEO title tag and meta description are update, could you take a look?

iogt/static/js/admin.js
```diff
@@ -21,6 +21,16 @@ $(document).ready(function () {
...
+
+    $('#id_seo_title').parent().siblings('p[class=help]').text(
+        'The name of the page displayed on search engine results as the clickable headline, and automatically loaded ' +
+        'as a title into shares on social media eg Facebook Posts or shared links in WhatsApp.'
+    );
+
+    $('#id_search_description').parent().siblings('p[class=help]').text(
+        'The descriptive text displayed underneath a headline in search engine results, and automatically loaded ' +
+        'as a description into shares on social media eg Facebook Posts or shared links in WhatsApp.'
+    );
```
istride, line 33: Let's take this opportunity to improve the organisation of this code. Currently, this is an anonymous function, but I think it would be better if it were given a name that summarises what it is doing - something like 'customizeUI'.

```
function customizeUI() {
...
}
$(document).ready(customizeUI);
```

istride (approved):
OK to merge after the suggested refactoring.

home/templates/home/tags/meta_tags.html
```diff
@@ -0,0 +1,19 @@
...
+    <meta property="og:site_name" content="{{ site_name }}">
+    <meta property="og:title" content="{% firstof page.seo_title page.title %}">
+    <meta property="og:description" content="{% firstof page.search_description site_name %}">
+    <meta property="og:type" content="article">
+    <meta property="twitter:site" content="{{ site_name }}">
+    <meta property="twitter:title" content="{% firstof page.seo_title page.title %}">
+    <meta property="twitter:description" content="{% firstof page.search_description site_name %}">
+    <meta property="twitter:card" content="summary_large_image">
+    {% if page.lead_image %}
+        {% image page.lead_image width-360 as lead_image %}
```
istride, line 14: This has been hard-coded to 'width-360' and I was expecting this to be based on `IMAGE_SIZE_PRESET` (full or half-width). I think it would be less surprising to use the same image sizes throughout, and for sites that use larger images there won't be a need to create extra image renditions just for social media previews.

istride (changes requested):

home/templates/home/tags/meta_tags.html
```diff
@@ -0,0 +1,18 @@
+{% load wagtailimages_tags %}
```
istride, line 1: not required?

home/templatetags/home_tags.py
```diff
@@ -158,3 +158,16 @@ def render_image(context, image, half_width=False, img_class=None):
         'class': img_class,
     })
     return context
+
+
+@register.inclusion_tag('home/tags/meta_tags.html', takes_context=True)
+def render_meta_tags(context):
```
istride, line 164: can we rename this to 'meta_tags' or 'social_meta_tags'? I want to describe a thing rather than an action.

home/templates/home/tags/meta_tags.html
```diff
@@ -0,0 +1,18 @@
...
+
+{% with site_name=page.get_site.site_name %}
```
istride, line 3: should we add site name to the context instead?

istride (changes requested):

PR #1579, opened 2023-03-15
Adds user activity tracking via Matomo

### LINKED ISSUE #1566: Add Javascript user activity tracking via Matomo
We can start the process of testing Matomo by adding simple JS tags.

We can consider this Django app: https://github.com/mikbuch/django-matomo

Alihassanc5:
Closes #1566 

- The feature adds tracking of user activity using Matomo


requirements.dev.txt
```diff
@@ -7,4 +7,5 @@ django-test-plus>=1.4.0
 coverage
 selenium==4.7.2
 factory-boy==3.2.*
-wagtail-factories==4.0.*
\ No newline at end of file
+wagtail-factories==4.0.*
+django-matomo
```
sheralim012, line 11: Specify version

sheralim012 (changes requested):

cbunicef:
Would it be possible to separately specify URLs for global and local Matomo tracking?

sheralim012:
> Would it be possible to separately specify URLs for global and local Matomo tracking?

It's doable but what's the use case behind this?

cbunicef:
> > Would it be possible to separately specify URLs for global and local Matomo tracking?
> 
> It's doable but what's the use case behind this?

For example, if we have an IoGT platform deployed by a UNICEF partner on their servers. They might use Matomo (local, cloud etc) to track locally, and we might want to use our global Matomo to also track globally.

We never needed to consider multiple systems for Google Analytics because there was only one system available, but since local deployments are possible with Matomo it would be something I could see us using.

This isn't urgent, but it seemed that since the team was currently working on this PR it would be faster now than digging into it again at a later date.

sheralim012 (changes requested):

cbunicef:
Please add a feature flag for this, while it's in development.

requirements.dev.txt
```diff
@@ -7,4 +7,5 @@ django-test-plus>=1.4.0
 coverage
 selenium==4.7.2
 factory-boy==3.2.*
-wagtail-factories==4.0.*
\ No newline at end of file
+wagtail-factories==4.0.*
+django-matomo==0.1.6
```
istride, line 11: django-matomo cannot be a dev-only dependency

istride (changes requested):

istride (approved):

PR #1596, opened 2023-04-26
Introduce cache key prefix configuration env var

istride:
At the moment, page caching (and other caching features) rely on having a single Redis server or cluster available for exclusive use by a single IoGT instance. In large deployments of IoGT, this situation would become quite time-consuming to manage. It would be useful to have the option to allow several instances of IoGT to share a single Redis instance.

The main problem is that Redis doesn't offer very robust options for sharing between multiple clients (apps). Redis does support multiple databases, but only up to 16, and Redis Cluster doesn't support multiple databases at all. The best that can be done is to partition the key space with a prefix, for each client that shares a Redis instance, to make sure that different clients don't interfere with one another. This partitioning method can be secured further by using Redis ACLs, but this is out of scope.

This PR introduces a new configuration environment variable that allows the key prefix to be set for a particular instance of IoGT. To make sure that the key prefix is used throughout the application the key prefix is configured via the Django Cache framework and it is expected that any component of the app that needs caching services should get them via the framework.

### DIFF
--- a/docs/cache.md
+++ b/docs/cache.md
@@ -1,20 +1,31 @@
 # Overview
 
-The page cache is provided by [Wagtail Cache][2] and is not enabled, by default. This guide leans towards Redis as the cache backend, but it should be possible to use any cache backend supported by Django 3.1.
+IoGT has performance features that can be enabled when a distributed cache, like Redis, is available, including:
+
+- A page cache, provided by [Wagtail Cache][2].
+- A cache for Wagtail image renditions
+- A general-purpose 'default' cache
+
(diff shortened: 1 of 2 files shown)


sheralim012 (approved):
LGTM 👍🏽,

One quick thing, if we are going to address the global data (translations, images, theme settings) caching separately then could you create a ticket for it because that wouldn't work with this?

istride:
@sheralim012 I think the 'global data' caching we already have can co-exist with page caching, but it won't be very effective by comparison. However, as it stands, the page caching does not work for logged in users, so it's probably worth keeping the 'global data' caching a bit longer. I'll assume that we'll keep things the way they are and if it becomes a problem we'll raise an issue for it. Sound good?

PR #1598, opened 2023-05-03
Integrate pip-tools for efficient package dependency management

### LINKED ISSUE #1618: Pip-tools for efficient package dependency management
The integration of Pip-tools offers several advantages, including automatic resolution of package version conflicts, easy management of pinned package versions, and the ability to generate hashed package requirements for improved security.

Alihassanc5:
Closes #1618


docs/packages.md
```diff
@@ -0,0 +1,15 @@
...
+```
+pip install --upgrade pip
+pip install pip-tools
+pip-compile --generate-hashes --resolver=backtracking -o <requirements_file>.txt <requirements_file>.in
+```
+
+### For Docker
+```
+make cr
+```
```
istride, line 15: I would like to improve the structure of this document.

- Rename to 'dependency-management.md', rather than 'packages.md'
- Introduce pip-tools: what is it; why do we use it, how do we use it?
- Ask readers to install the required tools (pip, pip-tools) as the first step, then amend the 'requirements.in' file
- Use headings appropriately, i.e. first-level, then second-level, then third-level, rather than straight to third-level (###)

istride (changes requested):

scripts/cr.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+REQ_FILES=(
+  "/app/requirements"
+  "/app/requirements.dev"
+)
+
+for f in "${REQ_FILES[@]}"; do
+  rm ${f}.txt && pip-compile --generate-hashes --resolver=backtracking -o ${f}.txt ${f}.in || exit 1;
```
istride, line 9: The `rm` command will fail if, for whatever reason (accident?), the current 'requirements' file doesn't already exist, and then the whole script will exit. We would not be able to compile a requirements file from scratch with this script, unless you trick the script by creating an empty requirements file first.

Is the `rm` command necessary? The pip-tools documentation suggest that `pip-compile` would behave sensibly if it found an existing requirements.txt file.

istride (approved):

PR #1608, opened 2023-05-12
Feat / Adds user tracking via Matomo

### LINKED ISSUE #1607: No Matomo analytics tracking on devices with no JavaScript support
Analytics tracking via Matomo was recently implemented in #1566, but this does not track devices with no JavaScript support. It is a hard requirement for IoGT to be able to be fully functional without JavaScript, including analytics tracking.

Alihassanc5:
Closes #1607

- Created New app called 'matomo' 
- This PR enables Matomo Analytics tracking for all devices that either JavaScript enabled or have JavaScript disabled.


iogt/settings/base.py
```diff
@@ -38,6 +38,7 @@
 INSTALLED_APPS = [
     'home',
     'search',
+    'iogt',
```
istride, line 41: This is unusual and I don't think we should be doing it. 'iogt' is a project, not an app.

iogt/templatetags/tracking_tags.py
```diff
@@ -0,0 +1,19 @@
...
+
+register = template.Library()
+
+
+@register.inclusion_tag('tracking_matomo_js_disabled_code.html')
+def tracking_matomo_js_disabled_code():
+    try:
+        id = settings.MATOMO_SITE_ID
+    except AttributeError:
+        raise ImproperlyConfigured('MATOMO_SITE_ID does not exist.')
```
istride, line 14: I know this is what the 'django-matomo' app does, but it's not useful or helpful for the site to blow up just because Matomo is not set up correctly - it is entirely optional. If we want to do this kind of thing, we should do it during startup, and take into account whether Matomo tracking is enabled or not.

iogt/templates/base.html
```diff
@@ -50,6 +50,7 @@
 {% get_current_language_bidi as LANGUAGE_BIDI %}
 <body class="{% if LANGUAGE_BIDI %}rtl{% endif %} {% block body_class %}{% endblock %}">
 {% include 'google_analytics_pixel_tracking.html' %}
+{% if matomo_tracking %} {% tracking_matomo_js_disabled_code %} {% endif %}
```
istride, line 53: I'd prefer this 'if' to be in the template for the tracking tag.

istride (changes requested):
The 'django-matomo' app project has just included the image beacon:

https://github.com/mikbuch/django-matomo/blob/master/matomo/templates/matomo/tracking_code.html

Not sure when the next release is going to be. In any case, 'django-matomo' is not doing very much at the moment, so I'm wondering whether we should just create our own matomo app that works just the way we want it to.

One thing to notice about 'django-matomo' is that they've included the image beacon next to the JavaScript tag.

Alihassanc5:
@istride, it seems prudent for us to utilize the code from the django-matomo branch. It aligns with our requirements for tracking on non-JavaScript devices, preventing us to repeat the code. What are your thoughts on this?

istride:
I think we should create our own app to replace django-matomo, which only provides a single template and templatetag that we would need to override anyway. This new app would need to provide:

- JavaScript tracking
- No script tracking ("Image Beacon")
- A global switch to activate/deactivate tracking
- A safe way to deal with incorrect configuration i.e. do nothing or log instead of throwing exceptions; maybe this makes a global switch redundant

Later, we may need to add the ability to manage configuration via site settings, and tracking app-specific events, so I think it makes sense to maintain our own app in preparation for those and other customisations in the future. I also think the door is still open to using django-matomo as a dependency, if it aligns with our use cases.

matomo/templates/matomo_tracking_code.html
```diff
@@ -0,0 +1,21 @@
+<!-- Matomo -->
+{% if matomo_tracking and url and site_id %}
+    Enable
```
istride, line 3: Clean up.

matomo/processors.py
```diff
@@ -0,0 +1,7 @@
+from django.conf import settings
+
+
+def matomo_tracking(request):
+    return {
+        'matomo_tracking': settings.MATOMO_TRACKING,
+    }
```
istride, line 7: Not getting much benefit from this processor since this context variable is only used in the tracking tag template. It would be better to move this to the templatetag.

matomo/templatetags/matomo_tags.py
```diff
@@ -0,0 +1,17 @@
...
+@register.inclusion_tag('matomo_tracking_code.html', takes_context=True)
+def matomo_tracking_code(context):
+    site_id = settings.MATOMO_SITE_ID
+    url = settings.MATOMO_URL
+
+    context.update({
+        'site_id': site_id,
+        'url': url
+    })
+    return context
```
istride, line 17: As mentioned elsewhere, I think the `matomo_tracking` variable should be added here too, instead of in the processor, however, with a couple of changes:

- rename key from 'matomo_tracking' to 'tracking_enabled'
- the value should consider the other required settings (site ID and server URL) i.e. tracking is only enabled if all the settings agree

We should be more explicit in the name of the 'url' variable and rename it to 'matomo_server_url', to make it very clear which URL and which _server_ URL we are referring to.

I'd like to rename the templatetag (and template) to 'matomo_tracking_tags'.

A unit test could be written for this templatetag, to make sure that 'tracking_enabled' is set correctly in the context. Are there any edge cases that could be problematic? The code is quite simple though, so I won't insist on it.

matomo/templatetags/matomo_tags.py
```diff
@@ -0,0 +1,15 @@
...
+from django.conf import settings
+
+
+register = template.Library()
+
+
+@register.inclusion_tag('matomo_tracking_code.html', takes_context=True)
+def matomo_tracking_tags(context):
+    context.update({
+        'tracking_enabled': settings.MATOMO_TRACKING,
```
istride, line 11: Please make this:
```
'tracking_enabled': (
    settings.MATOMO_TRACKING
    and settings.MATOMO_SERVER_URL
    and settings.MATOMO_SITE_ID
)
```

matomo/templatetags/matomo_tags.py
```diff
@@ -0,0 +1,15 @@
+from django import template
+from django.conf import settings
+
+
+register = template.Library()
+
+
+@register.inclusion_tag('matomo_tracking_code.html', takes_context=True)
```
istride, line 8: Please rename the template file to match the name of the templatetag.

istride (changes requested):

istride (approved):

PR #1609, opened 2023-05-12
Remove redundant test configuration

istride:
There is a lot of configuration settings in the various test configuration files that seem unnecessary, and removing them will allow us to see which settings _are_ important, without affecting the tests.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,37 +1,22 @@
 version: '3.5'
 
 services:
-
   django:
     build:
       context: ./
       dockerfile: Dockerfile.dev
     environment:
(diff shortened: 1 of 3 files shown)


docker-compose.test.yml
```diff
@@ -1,58 +1,26 @@
...
       - ./:/app/
-    ports:
-      - '8000:8000'
     depends_on:
-      - elasticsearch
       - db
-
   db:
     image: postgres:alpine
     restart: unless-stopped
```
sheralim012, line 22: I believe we can remove this as well.

sheralim012 (changes requested):
RLGTM 👍🏽

PR #1610, opened 2023-05-12
Reduce number of warning messages in test logs

istride:
There are repeated warnings about the BeautifulSoup HTML parser appearing in the logs while the unit tests are running. The warnings might obscure more important problems that may occur during test runs. The fix is simply to specify exactly which parser to use - I have set it to 'lxml'.

### DIFF
--- a/home/tests/tests.py
+++ b/home/tests/tests.py
@@ -1,13 +1,20 @@
+from bs4 import BeautifulSoup
 from django.test import TestCase, override_settings
 from django.http import HttpRequest
 from translation_manager.models import TranslationEntry
 from wagtail.core.models import Site
+from wagtail_factories import SiteFactory, PageFactory
 from wagtail_localize.operations import TranslationCreator
 
 from home.wagtail_hooks import limit_page_chooser
-from home.factories import SectionFactory, ArticleFactory, HomePageFactory, MediaFactory, LocaleFactory, SiteSettingsFactory
-from wagtail_factories import SiteFactory, PageFactory
-from bs4 import BeautifulSoup
+from home.factories import (
+    SectionFactory,
+    ArticleFactory,
+    HomePageFactory,
+    MediaFactory,
+    LocaleFactory,
+    SiteSettingsFactory
+)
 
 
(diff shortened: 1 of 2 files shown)


home/tests/tests.py
```diff
@@ -157,3 +164,7 @@ def test_half_custom_image_within_preset(self):
         self.assertEqual(int(rendered_image.get('width')), image_rendition.width)
         self.assertEqual(int(rendered_image.get('height')), image_rendition.height)
         self.assertEqual(rendered_image.get('src'), image_rendition.url)
+
+
+def parse_html(html: str) -> BeautifulSoup:
```
sheralim012, line 169: I know it's a one-liner function but I prefer to move this function to a more generic/shared place so other apps can utilize it without defining it as their own, which will keep it DRY and manageable.

sheralim012 (changes requested):
RLGTM 👍🏽

home/tests/tests.py
istride: I agree that this should be in a single, shared place, but I am not sure where that place is. The function is required by the 'questionnaires' and 'home' apps; do we keep it in 'home' and create a dependency on it in 'questionnaires'? Or, would you suggest something else?

sheralim012: `iogt/` project directory is the common place to hold these shared utils.

istride: Technically, that would work, but it would violate the principle of the project depending on the apps and not the other way around. The home and questionnaires apps would not be reusable outside of the iogt project. In theory, someone might want a version of IoGT without the questionnaire stuff.

I think questionnaires really can't exist without the home app, but home should be able to exist by itself, so that makes it a better location for shared code. A problem would arise if `parse_html` was needed by an app that doesn't currently depend on home, but perhaps we cross that bridge when we get to it.

istride: Resolved by removing `parse_html` in `questionnaires` and keeping it in `home`.

sheralim012 (approved):

PR #1611, opened 2023-05-15
Allow unit and selenium tests to run in parallel

istride:
This PR allows test to be run in several processes at the same time, which substantially reduces the amount of time required to run a full test suite - especially Selenium tests.

Several improvements were required to make this possible:
- Run Selenium tests against a PostgreSQL database instead of SQLite
- Allow several test servers to be started by the Selenium test suite
- Introduce an env var to control parallelism

Even if the tests are not run in parallel, the Selenium tests should be more stable now due to the use of PostgreSQL - tests were failing randomly due to not being able to flush the SQLite database.

### DIFF
--- a/docker-compose.selenium.yml
+++ b/docker-compose.selenium.yml
@@ -1,50 +1,32 @@
 version: '3.5'
 
 services:
-
   django:
     build:
       context: ./
       dockerfile: Dockerfile.dev
     environment:
-      DJANGO_SETTINGS_MODULE: iogt.settings.docker_compose_dev
-      COMMIT_HASH: asdfghjkl
-      WAGTAILTRANSFER_SECRET_KEY: wagtailtransfer-secret-key
-      WAGTAILTRANSFER_SOURCE_NAME: iogt_global
-      WAGTAILTRANSFER_SOURCE_BASE_URL: https://example.com/wagtail-transfer/
-      WAGTAILTRANSFER_SOURCE_SECRET_KEY: wagtailtransfer-source-secret-key
-      BASE_URL: 'http://localhost:8000'
-    command: python manage.py runserver 0.0.0.0:8000
+      DJANGO_SETTINGS_MODULE: iogt.settings.test
+    command: ["tini", "--", "sleep", "infinity"]
     volumes:
       - ./:/app/
-    ports:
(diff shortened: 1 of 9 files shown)


istride:
Parallelism is not active for Selenium tests running in Github Actions because the [standard runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) only has two CPU cores available to it and the Selenium tests already require about 5 different processes to run just so that the tests can be run sequentially.

docker-compose.selenium.yml
```diff
@@ -1,50 +1,32 @@
...
     depends_on:
-      - elasticsearch
+      - database
       - selenium-hub
-
-  elasticsearch:
-    image: 'docker.elastic.co/elasticsearch/elasticsearch:7.12.1'
+  database:
+    image: postgres:14-alpine
+    restart: unless-stopped
```
sheralim012, line 18: I think we can remove this.

selenium_tests/base.py
```diff
@@ -62,3 +53,56 @@ def setup_blank_site(self):
...
+        options.add_argument('--ignore-ssl-errors=yes')
+        options.add_argument('--ignore-certificate-errors')
+        options.add_argument("--window-size=480,720")
+        options.add_argument("--start-maximized")
+        driver = webdriver.Remote(
+            command_executor=get_hub_url(),
+            desired_capabilities=DesiredCapabilities.CHROME,
+            options=options,
+        )
+    elif preset == "firefox":
```
sheralim012, line 91: only the chrome node is available in the docker-compose, do we intend to add it later?

sheralim012 (changes requested):
RLGTM

selenium_tests/base.py
istride: I think it will have to be added later because some tests fail in Firefox, so more work is required to reconcile the differences between browsers. If a Firefox node is added to docker-compose now, it would not be used, so I've decided it's best to leave it out.

docker-compose.selenium.yml
istride: Removed.

sheralim012 (approved):

PR #1612, opened 2023-05-16
Add media URLs to Page Tree Api responses for offline access

### LINKED ISSUE #1379: Cannot play videos on the Offline app

istride:
Based on PR #1438. I thought it best to implement it myself to describe what I had in mind.

Closes #1379


home/models.py
```diff
@@ -1222,3 +1214,35 @@ def clean(self):
...
+    return [
+        img['src']
+        for img
+        in BeautifulSoup(str(text), 'lxml').find_all('img')
+    ]
+
+
+def collect_urls_from_streamfield(field):
+    urls = []
+    for block in field:
```
sheralim012, line 1237: my preference would be to traverse all the blocks and gather all the image ids and at the end pass it to `get_all_renditions_urls` as a list to reduce the number of DB queries.

iogt/views.py
```diff
@@ -121,10 +121,12 @@ def get(self, request, page_id):
...
         image_urls = []
+        media_urls = []
         for page in pages:
-            if isinstance(page, (HomePage, Section, Article, Poll, Survey, Quiz)):
-                page_urls.append(page.url)
-                image_urls += page.get_image_urls
+            try:
+                page_urls.append(page.offline_urls)
+            except AttributeError:
+                pass
```
sheralim012, line 129: Could you add a warning log here to describe the situation and on which page type it's occurring?

iogt/views.py
```diff
@@ -121,10 +121,12 @@ def get(self, request, page_id):
                 page_urls.append(page.url)
         translation.activate(active_locale.language_code)
         image_urls = []
+        media_urls = []
```
sheralim012, line 124: if not used, could you remove it?

home/models.py
```diff
@@ -1222,3 +1214,35 @@ def clean(self):
 
     def __str__(self):
         return f'{self.locale}'
+
+
+def get_all_renditions_urls(image):
```
sheralim012, line 1219: My preference would be to move these to `utils.py` or something like that to keep things organized.

sheralim012 (changes requested):

home/models.py
istride: Sounds good to me, however, there is already a `home.utils` module with sub-modules. Ideally, I would like to merge all the sub-modules into a single `home.utils` module, and then add these functions as well, but is that too much change?

An alternative would be to add an '__init__.py' in 'home/utils' and add the functions there - what do you think?

home/models.py
istride: I don't want to make a premature performance optimisation. This code is doing approximately the same as it was before, and I'm not sure if there was a performance issue before, but if there was, then perhaps it should be addressed separately.

Using a [renditions cache](https://docs.wagtail.org/en/stable/advanced_topics/performance.html#caching-image-renditions) would probably solve any performance issue, and we could set that up without Redis i.e. use the default in-memory cache, if we wanted to.

iogt/views.py
istride: good catch. removed.

iogt/views.py
istride: Added. Sample log, based on data from sandbox:

```
Offline URLs could not be obtained for the following Page types: BannerIndexPage, FooterIndexPage, MiscellaneousIndexPage, PageLinkPage, PollIndexPage, QuizIndexPage, SectionIndexPage, SurveyIndexPage
```

home/models.py
sheralim012: The first option sounds reasonable but one thing to consider in this approach is how big we want a module to be. If you think it's more manageable to combine utils then please go ahead ( my preference would be to do it in a separate PR to keep this PR relevant & to the point).

Another option would be to keep things in their module and import them in `home/utils/__init__.py` so other modules can import things from `home.utils` directly instead of each individual module.

istride: I'd to like to move towards a single utils module because there isn't enough code there to justify the existence of sub-modules, IMO. With that in mind, I have moved the functions from `home.models` to `home.utils`. I will create a separate PR to merge the sub-modules.

sheralim012 (approved):

PR #1658, opened 2023-12-07
Move link to comment into 'actions' section of CCM UI

istride:
To incorporate the link to comment more sympathetically with the rest of the Community Comment Moderation interface.


istride:
Screen Shot 2023-12-07 at 18 39 07

cbunicef:
For this I would rather have the link to the comment not sympathetic, actually, so that it is clear that the Actions are comment moderation actions. Could we keep the button, change the text to "View Comment", and put it centered underneath the text?

PR #1707, opened 2024-08-12
start interactive game by direct link

mizanriseup:
When any one browse direct interactive game link that will redirect to home page. We have fixed it here. Now if anyone hit or browse direct link of interactive game that will play the game.


istride:
Merged manually.

PR #1797, opened 2025-10-29
Feature/quiz inline feedback

Ankit-lab809:


sachin-mehta (commented):
1. Check date_of_birth, gender fields for existing users, if these are working fine or not.
2. The migration for QuizAttempt creation: ensure no conflict with existing data or absent foreign keys.
3. The PageVisit model may grow very large — consider if you need retention/archival strategy.
4. The CSS sets .activity-table { width:150% } – Might overflow container; check on various devices/screen sizes.
5. The CSS uses font-size: 0.9rem for buttons, line-height 1.2 – ensure readability and accessibility (WCAG).
6. Inline styling is present, move to css and ensure existing elements /selectors are not affected.

